PR #1, opened 2017-04-14
Making explicit prepare request / process response methods.

dhermes:
This is in line with the Sans-I/O philosophy.

### DIFF
--- a/gooresmed/download.py
+++ b/gooresmed/download.py
@@ -124,6 +124,41 @@ class Download(_DownloadBase):
            ``start`` to the end of the media.
     """
 
+    def _prepare_request(self):
+        """Prepare the contents of an HTTP request.
+
+        This is everything that must be done before a request that doesn't
+        require network I/O (or other I/O). This is based on the `sans-I/O`_
+        philosophy.
+
+        Returns:
+            dict: The headers for the request.
+
+        Raises:
+            ValueError: If the current :class:`Download` has already
+                finished.
+
+        .. _sans-I/O: https://sans-io.readthedocs.io/
+        """
+        if self.finished:
+            raise ValueError('A download can only be used once.')
+
+        headers = {}
+        _add_bytes_range(self.start, self.end, headers)
+        return headers
+
+    def _process_response(self):
+        """Process the response from an HTTP request.
+
+        This is everything that must be done after a request that doesn't
+        require network I/O (or other I/O). This is based on the `sans-I/O`_
+        philosophy.
+
+        .. _sans-I/O: https://sans-io.readthedocs.io/
+        """
+        # Tombstone the current Download so it cannot be used again.
+        self._finished = True
+
     def consume(self, transport):
         """Consume the resource to be downloaded.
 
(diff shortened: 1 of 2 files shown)


gooresmed/download.py
```diff
@@ -124,6 +124,41 @@ class Download(_DownloadBase):
...
+        Raises:
+            ValueError: If the current :class:`Download` has already
+                finished.
+        """
+        if self.finished:
+            raise ValueError('A download can only be used once.')
+
+        headers = {}
+        _add_bytes_range(self.start, self.end, headers)
+        return headers
```
dhermes: @jonparrott I'm a little "torn" here. I consider just returned all the inputs for the request:

```
pos_args = (self.media_url,)
kw_args = {'headers': headers}
return pos_args, kw_args
```

gooresmed/download.py
```diff
@@ -124,6 +124,41 @@ class Download(_DownloadBase):
            ``start`` to the end of the media.
     """
 
+    def _prepare_request(self):
+        """Prepare the contents of an HTTP request.
+
+        .. _sans-I/O: https://sans-io.readthedocs.io/
```
jonparrott: nit: I prefer link references to go at the end of the docstring.

gooresmed/download.py
jonparrott: I'd say you're welcome to return all of the args that are relevant or the complete set of args if you're going to make the transports conform to `google.auth.transport.Request`, either way, just be consistent.

gooresmed/download.py
```diff
@@ -124,6 +124,41 @@ class Download(_DownloadBase):
...
+                finished.
+        """
+        if self.finished:
+            raise ValueError('A download can only be used once.')
+
+        headers = {}
+        _add_bytes_range(self.start, self.end, headers)
+        return headers
+
+    def _process_response(self):
```
jonparrott: One thing you might think about is an abc for "HTTP Pipeline": https://github.com/haikuginger/httppipeline/blob/master/httppipeline/sample.py

gooresmed/download.py
```diff
@@ -124,6 +124,41 @@ class Download(_DownloadBase):
...
+        .. _sans-I/O: https://sans-io.readthedocs.io/
+
+        This is everything that must be done after a request that doesn't
+        require network I/O (or other I/O). This is based on the `sans-I/O`_
+        philosophy.
+        """
+        # Tombstone the current Download so it cannot be used again.
+        self._finished = True
+
     def consume(self, transport):
```
jonparrott: To be "truly" sans-io the part that does IO would need to be completely separate from the parts that prepare and process the IO. But, this is close enough for our purposes.

jonparrott: Nah, not that, just that they're different primitives.

e.g., you'd have `download.sansio` for all the sans-io pieces then `download.adapters` for the actual classes that implement 'consume'. 

It's up to you. IMO it's borderline overkill because this library will almost always be tied to google auth's transport interface.

dhermes: Seems overkill / overengineering to split the classes (let alone split the module) at this point.

jonparrott:
agreed

PR #2, opened 2017-04-19
Recover from error by resetting stream in ResumableUpload.

dhermes:
/cc @jonparrott


dhermes:
After discussing with @jonparrott, this fix doesn't make sense. This is because the upload will be in an unknown state and a ["resume"][1] request needs to be sent.

[1]: https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload#resume-upload

PR #3, opened 2017-04-19
Add ResumableUpload.invalid().

dhermes:
This way an instance can be marked when an unexpected response occurs.

### DIFF
--- a/gooresmed/upload.py
+++ b/gooresmed/upload.py
@@ -246,6 +246,16 @@ class ResumableUpload(_UploadBase):
         self._bytes_uploaded = 0
         self._total_bytes = None
         self._upload_url_with_id = None
+        self._invalid = False
+
+    @property
+    def invalid(self):
+        """bool: Indicates if the upload is in an invalid state.
+
+        This will occur if a call to :meth:`transmit_next_chunk` fails.
+        To recover from such a failure, call :meth:`recover`.
+        """
+        return self._invalid
 
     @property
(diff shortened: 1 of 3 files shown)


dhermes:
This is my suggested approach instead of #2, let me know WDYT reviewers

gooresmed/upload.py
```diff
@@ -417,14 +431,21 @@ def _process_response(self, response):
             # Tombstone the current upload so it cannot be used again.
             self._finished = True
         elif status_code == PERMANENT_REDIRECT:
-            bytes_range = _helpers.header_required(response, u'range')
+            try:
+                bytes_range = _helpers.header_required(response, u'range')
+            except exceptions.InvalidResponse:
```
dhermes: I don't quite like this part, but felt it was the "lesser of evil". **Every** other usage of `header_required` just raises (on failure) so I didn't think it'd be worth using a return value for raising.

jonparrott:
This seems reasonable enough to me.

dhermes:
@lukesneeringer I'm mergin with @jonparrott's

this-is-fine

but I **would** like some feedback if you have a chance to peak.

gooresmed/upload.py
dhermes: I "fixed it" by using a callback instead: https://github.com/GoogleCloudPlatform/google-resumable-media-python/commit/297387a12d9908ec194037e67120d0accafa4b55

PR #4, opened 2017-04-24
Adding retries for **ANY** http request.

dhermes:
Will always retry on 429, 500, 502, 503 and 504. Based on:
https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload#handling_errors

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -15,11 +15,36 @@
 """Shared utilities used by both downloads and uploads."""
 
 
+import random
+import time
+
+from six.moves import http_client
+
 from google.resumable_media import exceptions
 
 
 RANGE_HEADER = u'range'
 CONTENT_RANGE_HEADER = u'content-range'
+TOO_MANY_REQUESTS = 429
+"""int: Status code indicating rate-limiting.
+
+``http.client.TOO_MANY_REQUESTS`` was added in Python 3.3, so
+can't be used in a "general" code base.
+
+For more information, see `RFC 6585`_.
+
+.. _RFC 6585: https://tools.ietf.org/html/rfc6585#section-4
+"""
+RETRYABLE = (
+    TOO_MANY_REQUESTS,
+    http_client.INTERNAL_SERVER_ERROR,
+    http_client.BAD_GATEWAY,
+    http_client.SERVICE_UNAVAILABLE,
+    http_client.GATEWAY_TIMEOUT,
+)
+MAX_SLEEP = 64.0  # Just over 1 minute.
+MAX_SLEEP_EXPONENT = 6  # 2^6 = 64.
+MAX_CUMULATIVE_RETRY = 600.0  # 10 minutes
 
 
 def do_nothing():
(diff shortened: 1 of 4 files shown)


google/resumable_media/_helpers.py
```diff
@@ -131,4 +183,19 @@ def http_request(transport, method, url, data=None, headers=None):
...
         object: The return value of ``transport.request()``.
     """
-    return transport.request(method, url, data=data, headers=headers)
+    response = transport.request(method, url, data=data, headers=headers)
+    if get_status_code(response) not in RETRYABLE:
+        return response
+
+    total_sleep = 0.0
+    num_retries = 0
+    while total_sleep < MAX_CUMULATIVE_RETRY:
```
jonparrott: can you pull this behavior into a separate method/decorator/whatever?

dhermes: Not sure what you have in mind? Like a `do_stuff(func)` wrapper?

jonparrott: it seems like this retry logic is "unrelated" to the request object, so I'd prefer to pull it into its own thing, maybe something (conceptually) like:

```python
func = functools.partial(transport.request, method, url, data=data, headers)
predicate = lambda response: get_status_code(response) not in RETRYABLE
return _retry(func, predicate)
``

**Comment by dhermes** [comment]
@jonparrott Requested re-factor has been done. PTAL

File: google/resumable_media/_helpers.py [hunk 2]
```diff
@@ -114,6 +140,81 @@ def require_status_code(response, status_codes, callback=do_nothing):
...
+    Success is determined by ``predicate(result)`` being true.
+
+    Will retry until :attr:`MAX_CUMULATIVE_RETRY` seconds of wait time
+    have accrued. Uses :func:`calculate_retry_wait` to double the
+    wait time (with jitter) after each attempt.
+
+    Args:
+        func (Callable): A callable that takes no arguments and produces
+            a result which will be checked by ``predicate``.
+        predicate (Callable): Determines if the result is valid.
```
jonparrott: Callable[Any, bool]

google/resumable_media/_helpers.py
```diff
@@ -114,6 +140,81 @@ def require_status_code(response, status_codes, callback=do_nothing):
...
+            return result
+
+    return result
+
+
+def not_retryable_predicate(response):
+    """Determines if a ``response`` is retryable.
+
+    Args:
+        object: The return value of ``transport.request()``.
```
jonparrott: This is always `google.auth.transport.Response`.

google/resumable_media/_helpers.py
dhermes: Done (and folded into the re-factor commit)

google/resumable_media/_helpers.py
dhermes: The idea is that `transport` can be anything, which is why I'm strongly considering making `google.resumable_media.FOO` subpackages so it is **ALWAYS** explicit

PR #6, opened 2017-05-01
Adding support for resumable uploads of unknown size.

dhermes:


docs/latest/google.resumable_media.requests.html
```diff
@@ -808,8 +808,17 @@
```
jonparrott: weird that we have compiled docs in the source control?

dhermes: I preferred this over a `gh-pages` branch:

screenshot from 2017-05-01 09-35-07

PR #11, opened 2017-06-22
Adding docs updates after latest version of Sphinx.

dhermes:


dhermes:
These were generated just by running `nox -s docs` locally. Auto-merging here.

PR #14, opened 2017-07-18
Using nox `session.skip` (instead of ValueError) for system tests.

dhermes:


dhermes:
Merging this without review. (Please don't be angry reviewers, it's "not important" and it's a non-controversial config-only change that is needed for green PR builds.)

Issue #5, opened 2017-04-24
Should stream / iter_contents be used when using `Download`


dhermes:
The primary motivation is that users could avoid putting the contents into RAM all at once and just flush right to disk ([Context][1]). `requests` has support for a ["lazy" response][2] body.

/cc @lukesneeringer @jonparrott @tseaver 

----

It's worth noting that I have a goal of making `google.resumable_media.FOO` subpackages that **explicitly** support a given transport (mentioned in [first release notes][3]). Once something like this exists, it makes more sense to add in transport-specific optimizations.

[1]: https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3323#discussion_r113034659
[2]: http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow
[3]: https://github.com/GoogleCloudPlatform/google-resumable-media-python/releases/tag/0.0.1

PR #20, opened 2017-07-31
Adding optional stream argument to DownloadBase.

dhermes:
This is towards a fix for #5 but I wanted to discuss before doing the more "serious" work.

/cc @tseaver who I can't assign


google/resumable_media/_download.py
```diff
@@ -162,10 +185,15 @@ def _process_response(self, response):
         self._finished = True
         _helpers.require_status_code(
             response, _ACCEPTABLE_STATUS_CODES, self._get_status_code)
+        if self._stream is not None:
+            self._write_to_stream(response)
```
dhermes: This is an iffy proposition (at least as far as #5 is concerned).

`_process_response` is meant to be sans-I/O, but we'd like `requests.Download._write_to_stream` to override the default behavior and consume the content in chunks (still a single GET, just HTTP chunking). But this would be network I/O, so would belong in `consume()`, not here.

dhermes: The alternative is just that `Download._stream` is only really used by the implementation classes (`requests.Download` is the only one that exists now and may be the only one that ever exists)

jonparrott: It's fine for this to indirectly perform network IO, just as long as this class provides some means of replacing the network IO in subclasses.

tseaver: AFAIK, "too much stuff to keep in RAM" is the Achilles' heel of "sans IO".

dhermes: I'm not sure I even know what that'd like look. Instead of tilting at windmills I'm going to rip this out and make this PR only add the `_stream` attribute.

dhermes: @tseaver Sorry I missed your comment when I made mine.

I am (for now) just punting on figuring out what the sans-I/O approach should be. I will be following this up with modifying `requests.Download.consume()` to do the HTTP chunking in the "I am doing I/O now" section.

dhermes:
Will merge when this goes green. I'm assuming @jonparrott's LGTM extends to this PR with even less code.

PR #21, opened 2017-07-31
Using requests `iter_content()` to avoid storing response body in RAM.

### LINKED ISSUE #5: Should stream / iter_contents be used when using `Download`

dhermes:
Fixes #5.

I'll be adding a system test to make extra double sure this code works as expected, but wanted to send out the PR before adding it.

/cc @tseaver since I can't add you as a reviewer


google/resumable_media/requests/download.py
```diff
@@ -19,6 +19,9 @@
 from google.resumable_media.requests import _helpers
 
 
+_SINGLE_GET_CHUNK_SIZE = 8192
```
dhermes: I'd love some guidance here on what makes sense, maybe @thobrla has thought about this?

google/resumable_media/requests/_helpers.py
```diff
@@ -88,11 +88,14 @@ def http_request(transport, method, url, data=None, headers=None,
...
+        transport_kwargs (Dict[str, str]): Extra keyword arguments to be
+            passed along to ``transport.request``.
 
     Returns:
         ~requests.Response: The return value of ``transport.request()``.
     """
     func = functools.partial(
-        transport.request, method, url, data=data, headers=headers)
+        transport.request, method, url, data=data, headers=headers,
+        **transport_kwargs)
```
dhermes: @jonparrott I just realized that allowing `stream=True` to go through here isn't a free lunch, because `_helpers.wait_and_retry()` expects to be retrying a failure for a single request, but ISTM the socket might fail midway through?

Any ideas how we could mock and test this with `urllib3` / `requests`?

google/resumable_media/requests/download.py
dhermes: @thorbla, if you do make it here, some context:

This change allows use of [`iter_content()`][1] with `requests` to avoid having to read the entire file into RAM before writing it to disk.

[1]: http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow

thobrla: Does this translate into an HTTP request boundary or does it just represent how much data we're pulling off the wire at once?

dhermes: The latter: "how much data we're pulling off the wire at once"

dhermes:
I pushed the branch to the main project remote so the system tests would run: https://circleci.com/gh/GoogleCloudPlatform/google-resumable-media-python/176

google/resumable_media/requests/download.py
thobrla: 8192 is what gsutil uses, so it should be fine (assuming the underlying library pulling the data of the wire does this in a tight loop).  However, I haven't done an in-depth investigation/optimization of this constant, so it's possible there is room for improvement.

dhermes: SG, thanks! It's funny we ended up with the same constant. I was just winging it, but figured 8K was better than 1K!

google/resumable_media/requests/_helpers.py
jonparrott: Mock at the socket level. ;)

dhermes:
This may not be the end of the road, I am currently playing around with failures after consuming part of the `GET` payload.

PR #23, opened 2017-08-01
Updating version to 0.2.2.

dhermes:
Release notes:

- Added optional `stream` argument to `requests.Download`. If uses, the downloaded resource will be written directly to the `stream` in 8KB chunks, rather than being read into RAM. (#20, #21)

/cc @tseaver 

----

I spent about an hour this morning trying to get a "read" of one of those 8K chunks to fail, so we could figure out what retries might look like, but I had no luck.


dhermes:
OK I am dumb, the call to `_process_response` needs to happen before writing to the file because it actually verifies the response is valid. DERP!

Luckily I killed the [build][1] before `0.2.2` was pushed, so I can just "paper-over" that release.

[1]: https://circleci.com/gh/GoogleCloudPlatform/google-resumable-media-python/180

PR #24, opened 2017-08-01
Swap the order of _write_to_stream() / _process_response() in `requests` download.

dhermes:
Since this doesn't actually change the interface, none of the docs for `latest/` or `0.2.2/` need be changed.

Once merged, I propose this commit be tagged as the NEW-AND-IMPROVED `0.2.2` release

/cc @tseaver


tseaver:
LGTM, including moving the `0.2.2` tag (since PyPI still has `0.2.1` as the latest release).

PR #25, opened 2017-08-07
Allow empty files to be uploaded.

lukesneeringer:
@dhermes I am not sure this is actually the correct fix for GoogleCloudPlatform/google-cloud-python#3685, but it _seems_ logical and in line with what the original reporter expected.

If this is the wrong answer, I would appreciate guidance on what the correct answer is.

Fixes GoogleCloudPlatform/google-cloud-python#3685

### DIFF
--- a/google/resumable_media/_upload.py
+++ b/google/resumable_media/_upload.py
@@ -526,7 +526,7 @@ class ResumableUpload(UploadBase):
                 u'This upload has not been initiated. Please call '
                 u'initiate() before beginning to transmit chunks.')
 
-        start_byte, end_byte, payload, content_range = get_next_chunk(
+        start_byte, payload, content_range = get_next_chunk(
             self._stream, self._chunk_size, self._total_bytes)
         if start_byte != self.bytes_uploaded:
             msg = _STREAM_ERROR_TEMPLATE.format(
@@ -784,14 +784,15 @@ def get_next_chunk(stream, chunk_size, total_bytes):
             in the ``stream``.
 
     Returns:
-        Tuple[int, int, bytes, str]: Quadruple of:
+        Tuple[int, bytes, str]: Triple of:
 
           * the start byte index
-          * the end byte index
           * the content in between the start and end bytes (inclusive)
           * content range header for the chunk (slice) that has been read
 
     Raises:
+        ValueError: If ``total_bytes == 0`` but ``stream.read()`` yields
+            non-empty content.
         ValueError: If there is no data left to consume. This corresponds
             exactly to the case ``end_byte < start_byte``, which can only
             occur if ``end_byte == start_byte - 1``.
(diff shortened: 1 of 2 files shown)


google/resumable_media/_upload.py
```diff
@@ -817,7 +817,7 @@ def get_next_chunk(stream, chunk_size, total_bytes):
             raise ValueError(msg)
 
     content_range = get_content_range(start_byte, end_byte, total_bytes)
-    return start_byte, end_byte, payload, content_range
+    return start_byte, max([end_byte, 0]), payload, content_range
```
dhermes: @lukesneeringer This feels a little hacky. Let me dig in to the source and see if there is a better place for it.

dhermes: Also: I have confirmed that an empty upload works just fine with `SimpleUpload`

dhermes: ...but that won't help anyone using `google-cloud-storage` because it uses `MultipartUpload` and `ResumableUpload`

dhermes: Also just confirmed empty bytes works just fine for `MultipartUpload`.

dhermes: Also just confirmed this WON'T happen with `Blob.upload_from_file(..., size=None)` because it sets `stream_final=False`, so `ResumableUpload._total_bytes is None` (whereas `Blob.upload_from_filename` explicitly passes `size=os.fstat(file_obj.fileno()).st_size` to `upload_from_file`)

google/resumable_media/_upload.py
```diff
@@ -807,6 +808,13 @@ def get_next_chunk(stream, chunk_size, total_bytes):
         if num_bytes_read < chunk_size:
             # We now **KNOW** the total number of bytes.
             total_bytes = end_byte + 1
+    elif total_bytes == 0:
+        # NOTE: We also expect ``start_byte == 0`` here but don't check
+        #       because ``_prepare_initiate_request()`` requires the
+        #       stream to be at the beginning.
+        if num_bytes_read != 0:
+            raise ValueError(
+                u'Stream specified as empty, but produced non-empty content.')
```
dhermes: @lukesneeringer PTAL.

In either the `total_bytes == 0` and `total_bytes is None` case, we end up with:

```python
>>> get_content_range(0, -1, 0)
u'bytes */0'
```

If we "corrected" the value, we'd instead have:

```python
>>> get_content_range(0, 0, 0)
u'bytes 0-0/0'
```

IMO it's not worth correcting, LMK what you think

lukesneeringer: I was not sure if `-1` had negative side effects. I trust you to know. :-)

dhermes:
Merging based on a Hangouts LGTM from @lukesneeringer

google/resumable_media/_upload.py
dhermes: Indeed, this was the only side effect.

PR #26, opened 2017-08-07
Updating version to 0.2.3.

dhermes:
Release notes:

- Allow empty files to be sent in `ResumableUpload` of known size (i.e. when `initiate(..., stream_final=True)` or `initiate(..., total_bytes=0)`). Previously failed with "Stream is already exhausted..." (#25)

/cc @tseaver


dhermes:
https://circleci.com/gh/GoogleCloudPlatform/google-resumable-media-python/189

PR #31, opened 2017-10-06
Add checksum validation for non-chunked non-composite downloads

mfschwartz:
This partially addresses https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/22

Note current limitations:
  - doesn't handle chunked download case
  - doesn't handle case of downloading composite objects
  - doesn't do anything about upload checksumming

I ran the unit and system tests only against Python 2.7, because my current OS doesn't support Python 3.6.

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -16,7 +16,7 @@
 
 import base64
 import hashlib
-import logging
+import sys
 
 from google.resumable_media import _download
 from google.resumable_media.common import DataCorruption
@@ -68,11 +68,6 @@ class Download(_helpers.RequestsMixin, _download.Download):
                 server-computed checksum.
         """
         md5_hash = hashlib.md5()
-        # TODO: Add support for validating CRC32C for composite objects. Note
-        # that this is somewhat painful in Python 2 because most OS distros of
-        # Python 2 don't include a compiled crcmod (and executing without a
-        # compiled crcmod is very slow; and installing a compiled crcmod will
-        # make getting google-cloud-python working well more difficult).
         expected_md5_hash = None
         if ('X-Goog-Hash' in response.headers
             and response.headers['X-Goog-Hash']):
@@ -81,7 +76,7 @@ class Download(_helpers.RequestsMixin, _download.Download):
                 if name == 'md5':
                     expected_md5_hash = value
         if not expected_md5_hash:
-            logging.getLogger().info(
+            self._logger.info(
                 'No MD5 checksum was returned from the service while '
                 'downloading %s (which happens for composite objects), so '
                 'client-side content integrity checking is not being '
@@ -92,7 +87,12 @@ class Download(_helpers.RequestsMixin, _download.Download):
             for chunk in body_iter:
                 self._stream.write(chunk)
                 md5_hash.update(chunk)
-        actual_md5_hash = base64.encodestring(md5_hash.digest()).rstrip('\n')
+        if sys.version_info[:3] < (3,):
+            actual_md5_hash = (base64.encodestring(md5_hash.digest())
+                               .rstrip('\n'))
+        else:
+            actual_md5_hash = (base64.encodebytes(bytes(md5_hash.digest(),
+                                                        'UTF-8')).rstrip('\n'))
         if expected_md5_hash and actual_md5_hash != expected_md5_hash:
               raise DataCorruption(response,
                                    'Checksum mismatch while downloading %s: '
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -14,8 +14,12 @@
 
 """Support for downloading media from Google APIs."""
 
+import base64
+import hashlib
+import logging
 
 from google.resumable_media import _download
+from google.resumable_media.common import DataCorruption
 from google.resumable_media.requests import _helpers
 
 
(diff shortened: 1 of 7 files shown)


google/resumable_media/requests/download.py
```diff
@@ -58,12 +62,43 @@ def _write_to_stream(self, response):
 
         Args:
             response (~requests.Response): The HTTP response object.
+
+        Raises:
+            DataCorruption: If the download's checksum doesn't agree with
+                server-computed checksum.
         """
+        md5_hash = hashlib.md5()
+        # TODO: Add support for validating CRC32C for composite objects. Note
```
dhermes: @mfschwartz Please remove the `TODO` from the source and put it in the issue tracker.

google/resumable_media/requests/download.py
```diff
@@ -58,12 +62,43 @@ def _write_to_stream(self, response):
...
+        # that this is somewhat painful in Python 2 because most OS distros of
+        # Python 2 don't include a compiled crcmod (and executing without a
+        # compiled crcmod is very slow; and installing a compiled crcmod will
+        # make getting google-cloud-python working well more difficult).
+        expected_md5_hash = None
+        if ('X-Goog-Hash' in response.headers
+            and response.headers['X-Goog-Hash']):
```
dhermes: First, this should go in the base module `google.resumable_media._download` rather than the `requests`-specific implementation.

----

I suggest instead:

```python
headers = self._get_headers(response)
goog_hash = headers.get('X-Goog-Hash', '')
if goog_hash:
    ...
```

Are you worried about `headers` being case-sensitive (it won't be for `requests`)?

google/resumable_media/requests/download.py
```diff
@@ -58,12 +62,43 @@ def _write_to_stream(self, response):
...
+            for checksum in response.headers['X-Goog-Hash'].split(','):
+                name, value = checksum.split('=', 1)
+                if name == 'md5':
+                    expected_md5_hash = value
+        if not expected_md5_hash:
+            logging.getLogger().info(
```
dhermes: Instead we should make this logger a non-public attribute of the module.

setup.py
```diff
@@ -34,7 +34,7 @@
 
 setuptools.setup(
     name='google-resumable-media',
-    version='0.2.3',
+    version='0.2.4',
```
dhermes: Let's hold off on a revision / release until another PR dedicated to that purpose (I can handle that).

google/resumable_media/requests/download.py
mfschwartz: google.resumable_media._download doesn't have code that walks through the content chunk-by-chunk. So, for example, if I put the MD5 checksumming in _process_response it would either get the entire body at once (using self._get_body(response)) or I would have to add another loop over response.iter_content that walks through the chunks. The former would require too much memory for large downloads, and the latter would mean we're walking over the content twice (once in google.resumable_media._download, and once in google.resumable_media.requests.download, in _write_to_stream) - and I think the latter would fail because iter_content walks through the HTTP stream, and it wouldn't be available twice (well, not without starting a separate connection to fetch the content again). Both approaches seem much worse than what I have right now, but maybe I'm not understanding something about what you suggest?

google/resumable_media/requests/download.py
mfschwartz: Done

google/resumable_media/requests/download.py
mfschwartz: Done

setup.py
mfschwartz: Done

google/resumable_media/requests/download.py
```diff
@@ -58,12 +62,43 @@ def _write_to_stream(self, response):
...
                 chunk_size=_SINGLE_GET_CHUNK_SIZE, decode_unicode=False)
             for chunk in body_iter:
                 self._stream.write(chunk)
+                md5_hash.update(chunk)
+        if sys.version_info[:3] < (3,):
+            actual_md5_hash = (base64.encodestring(md5_hash.digest())
+                               .rstrip('\n'))
+        else:
+            actual_md5_hash = (base64.encodebytes(bytes(md5_hash.digest(),
+                                                        'UTF-8')).rstrip('\n'))
```
mfschwartz: FYI I added this conditional logic because of the test failure at https://circleci.com/gh/GoogleCloudPlatform/google-resumable-media-python/195#build-timing/containers/0
(I am unable to install Python3.6 on my machine, so Luke suggested just testing under Python 2.7 and letting Travis CI do the 3.6 tests...)

mfschwartz:
Hi - I'd appreciate if we could draw this review to a close soon because the lack of checksumming support in this library is causing customer problems. Thanks.

lukesneeringer:
> @mfschwartz I am happy to maintainer-ify this PR if you are OK with it? I think it may be easier than a nit-picky review.

Yes please.

dhermes:
@mfschwartz I'm closing this PR in favor of #32 (which uses your 3 commits as a diffbase).

Usually GitHub allows maintainers to push into a PR branch, but I couldn't for some reason (possibly because you used the `master` branch).

As for not being able to install Python 3 on your machine, if you are on Mac OS X or Linux, I highly recommend `pyenv` (it does not require root privileges). If you are on Windows, I recommend `conda` or Miniconda (ditto on admin privileges). I'm happy to help if it's relevant to you. If not, no worries.

dhermes:
> ... the lack of checksumming support in this library is causing customer problems

If you are referring to https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4129, it seems that the checksumming is a **theory** for that breakage, but we have not been able to reproduce the customer's report.

mfschwartz:
Thanks for the commits Danny.

The problem I was referring to was not GoogleCloudPlatform/google-cloud-python#4129, but rather a customer who reached out to the GCS team directly.

mfschwartz:
Can you please tell me when you have the two libraries in a state where it's possible to build new releases (at a min I think that means updated version #s and dependencies)? Thanks!

dhermes:
Yes. I'm worried that `google-cloud-core` will make releasing painful. Hopefully not.

dhermes:
@mfschwartz Am I correct in assuming you are OK with a release with the partial support implemented?

mfschwartz:
Yes, a release with partial support would be good.
I'm sure you're going to but can you please make sure all the tests pass? I tried to fix the test failure on python 3.6 but since I don't currently have a python 3.6 installed I can't test it.
(I saw your note about installing 3.6 on various platforms. I could do that on my Mac; I just did this development on an Ubuntu version that currently doesn't yet support 3.6.)

dhermes:
Yes, they are all passing: https://circleci.com/gh/GoogleCloudPlatform/google-resumable-media-python/198

----

My primary dev machine is Ubuntu and I use `pyenv` and never looked back. It's as easy as:

```
git clone https://github.com/pyenv/pyenv.git ~/.pyenv
~/.pyenv/bin/pyenv install 3.6.3
~/.pyenv/versions/3.6.3/bin/python3.6
```

If you want to just type `python3.6`, check out the README on the GitHub repo for things to add to your `bashrc` (or `zshrc`, etc.)

mfschwartz:
Ah, thanks, that was quite easy!

PR #35, opened 2017-10-20
Adding a `gzip`-ed text file for requests download system test.

dhermes:
This has revealed at least two errors and this is intended to be checked in in a broken state, so that subsequent PRs can fix it. (See https://github.com/dhermes/google-resumable-media-python/commit/29617b6fa0c4d3275dd54b42a78a9c0ed99cb2ce for a preview of one such fix).

In order to make this easy, I re-factored the tests in a commit before, so it's likely easier to review this PR commit-by-commit.

/cc @tseaver @mfschwartz


tests/system/requests/test_download.py
```diff
@@ -31,11 +32,41 @@
...
+            slice(-256, None, None),  # obj[-256:]
+            slice(262144, None, None),  # obj[262144:]
+        ),
+    }, {
+        u'path': os.path.realpath(os.path.join(DATA_DIR, u'file.txt.gz')),
+        u'uncompressed': os.path.realpath(os.path.join(DATA_DIR, u'file.txt')),
+        u'content_type': PLAIN_TEXT,
+        # NOTE: This **should** be u'KHRs/+ZSrc/FuuR4qz/PZQ=='.
+        u'checksum': u'XHSHAr/SpIeZtZbjgQ4nGw==',
+        u'slices': (),
```
dhermes: This is intentionally empty for now, but should be populated at a later time.

dhermes:
Thanks @tseaver!

PR #36, opened 2017-10-20
Add requests/urllib3 work-around for intercepting gzipped bytes.

### LINKED ISSUE #34: MD5 validation broken?
I have a file that I uploaded and then downloaded with `google-resumable-media` under the hood. It seems that 0.3 reports that the `X-Goog-Hash` header reports a different MD5 than the computed one. The downloaded file however is intact.

The only thing is my file is gzipped. Does gzip break the MD5 calculation?

Thanks!

----

**Also to be closed**: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4227

dhermes:
Fixes #34.

/cc @tseaver @mfschwartz

@jonparrott Do you think we should cc the `requests` / `urllib3` maintainers? I'm also working on a PR to fix the failing system test, but it is **not** going to be part of this PR.

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -18,6 +18,8 @@ import base64
 import hashlib
 import logging
 
+import urllib3.response
+
 from google.resumable_media import _download
 from google.resumable_media import common
 from google.resumable_media.requests import _helpers
@@ -113,11 +115,14 @@ class Download(_helpers.RequestsMixin, _download.Download):
         else:
             md5_hash = hashlib.md5()
         with response:
+            # NOTE: This might "donate" ``md5_hash`` to the decoder and replace
+            #       it with a ``_DoNothingHash``.
+            local_hash = _add_decoder(response.raw, md5_hash)
             body_iter = response.iter_content(
                 chunk_size=_SINGLE_GET_CHUNK_SIZE, decode_unicode=False)
             for chunk in body_iter:
                 self._stream.write(chunk)
-                md5_hash.update(chunk)
+                local_hash.update(chunk)
 
         if expected_md5_hash is None:
             return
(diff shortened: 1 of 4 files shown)


jonparrott:
Go for it.

google/resumable_media/requests/download.py
```diff
@@ -18,6 +18,8 @@
 import hashlib
 import logging
 
+import urllib3.response
```
dhermes: @jonparrott This makes it impossible to import this module unless `urllib3` is installed. Previously, `requests` could be uninstalled and everything would work perfectly if the passed in objects matched the expected interfaces.

I can move back to that approach and ditch the subclass. It's a [small implementation][1]:

```python
class GzipDecoder(object):

    def __init__(self):
        self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)

    def __getattr__(self, name):
        return getattr(self._obj, name)

    def decompress(self, data):
        if not data:
            return data
        return self._obj.decompress(data)
```

[1]: https://github.com/shazow/urllib3/blob/ff5c6749c237ff6a8b0a3450713ecce7469a7ea8/urllib3/response.py#L55-L66

mfschwartz:
I would suggest that rather than just Cc'ing the `requests / urllib3` maintainers, we recommend adding a feature to their libraries so the caller can get ahold of the pre-decoded content to do things like computing checksums. I think that capability has more general value than just for this library.

google/resumable_media/requests/download.py
```diff
@@ -286,3 +291,58 @@ def update(self, unused_chunk):
...
+        md5_hash (Union[_DoNothingHash, hashlib.md5]): A hash function which
+            will get updated when it encounters compressed bytes.
+
+    Returns:
+        Union[_DoNothingHash, hashlib.md5]: Either the original ``md5_hash``
+        if ``_decoder`` is not patched. Otherwise, returns a ``_DoNothingHash``
+        since the caller will no longer need to hash to decoded bytes.
+    """
+    encoding = response_raw.headers.get(u'content-encoding', u'').lower()
+    if encoding != u'gzip':
```
mfschwartz: ISTR that the accept-encoding is set to "gzip, deflate".
I think if the server decides to send deflate content the code here won't work, and instead will need to handle both cases.

dhermes:
@Lukasa @shazow We're curious what you think of this hack.

For Google Cloud Storage (and likely other media APIs with checksumming) we "need" simultaneous access to the (gzip) compressed and uncompressed bytes so we can compute the checksum of the gzipped content (which is what is actually stored in "the cloud") while streaming the uncompressed bytes to a stream for the user.

google/resumable_media/requests/download.py
dhermes: I based this off the code [from `urllib3`][1]. I can just change it to

```python
if u'gzip' not in encoding:
    ...
```

However, `gzip, deflate` is a **list** of encodings intended for `Accept-Encoding`. This is the `Content-Encoding` of the bytes.

[1]: https://github.com/shazow/urllib3/blob/ff5c6749c237ff6a8b0a3450713ecce7469a7ea8/urllib3/response.py#L256-L258

mfschwartz: Right, but if the server sends data compressed using deflate encoding, the current code won't monkey patch the decoder to handle that case, which means we'll try to compare the checksum with the Requests-decoded data and get a checksum failure.

dhermes: This PR is **explicitly** about `gzip`. If you think we need to handle the `deflate` encoding, that's a separate discussion (and this is the first I'm hearing it).

mfschwartz: Yes, I didn't think of it before.

I believe everything will work right now because if you make a request to GCS with accept-encoding: 'gzip, deflate' GCS happens to send using gzip; but it would be spec-legal to send in deflate format instead, and that would cause the library to break in the way I noted above.

A simple soln would be to remove 'deflate' in the accept-encoding for now.
(Side note: I've read that deflate is somewhat more CPU efficient than gzip, which I think is more of a concern for browsers in general.)

dhermes: @mfschwartz Can you file an issue that we can follow up on? This change needs to go in soon so we can release a `0.3.1` and un-break at least **some subset** of people.

mfschwartz:
Opened https://github.com/GoogleCloudPlatform/google-resumable-media-python/issues/37 about the deflate encoding issue noted above.

shazow:
@dhermes Looks like you did what you could with the API we have. :) It really does feel like the decoder stuff should be a registry, doesn't it?

PR #40, opened 2017-10-20
Adding AppVeyor support.

dhermes:
I'm doing this in **anticipation** of adding CRCMOD support (i.e. a C dependency).

We want to know that the extension can build on Windows and likely want to just get wheels directly from the builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,59 @@
+version: 1.0.{build}.{branch}
+
+build: off
+
+matrix:
+  fast_finish: true
+
+# We always use a 64-bit machine, but can build x86 distributions
+# with the PYTHON_ARCH variable.
+platform:
+    - x64
+
+environment:
+
+  matrix:
+
+    # See: https://www.appveyor.com/docs/installed-software/#python
+
+    - PYTHON: "C:\\Python27"
+    - PYTHON: "C:\\Python27-x64"
+    - PYTHON: "C:\\Python34"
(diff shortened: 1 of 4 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,59 @@
...
+  - python -m pip install --upgrade pip setuptools
+  - python -m pip install --upgrade virtualenv wheel
+
+  # Set up virtual environment for testing.
+  - python -m virtualenv test-venv
+  - test-venv\Scripts\pip install --upgrade --requirement appveyor\requirements.txt
+
+build_script:
+  # Build an actual wheel for the source in current form.
+  - python setup.py bdist_wheel --dist-dir dist
```
dhermes: After this I plan on uploading the wheel as an artifact if [`APPVEYOR_REPO_TAG`][1] is `true`. Something like:

```yaml
    - ps: |
        # Upload artifact to Appveyor immediately after build
        ls dist -r | Foreach-Object {
            appveyor PushArtifact $_.FullName
        }
```

(I've been around this block a few times.)

[1]: https://www.appveyor.com/docs/environment-variables/

dhermes:
https://ci.appveyor.com/project/dhermes/google-resumable-media-python/build/1.0.1.master is the build.

@jonparrott @lukesneeringer I **used** to be able to create projects under `GoogleCloudPlatform` on AppVeyor but it seems I've lost that ability. Could one of you create that project?

dhermes:
I will merge this as-is since it isn't controversial.

Issue #43, opened 2017-11-03
urllib3 dependency


brandonw:
Commit b6c62d80ad8415626fade6caf4282939862996bd introduced a dependency on the urllib3 library without a corresponding addition in the setup.py

As of now, if the latest version of google-resumable-media-python is used in a project that does not otherwise require urllib3, then the following error will be encountered:

`ImportError: No module named urllib3.response`

dhermes:
@brandonw It's included transitively by the [`requests >= 2.18.0`][1] "extra" and the `google.resumable_media.requests` package can't be used without the extra.

How are you installing `google-resumable-media`? If you've done it via `google-cloud-storage`, you will have gotten the extra.

Also, see #42 for a previous discussion of this. I am going to close since this is "working as intended" but am happy to continue the discussion.

[1]: https://github.com/GoogleCloudPlatform/google-resumable-media-python/blob/8700f14e7c72501df557aee95c8a2426dc8c4f4c/setup.py#L31

brandonw:
@dhermes Ah, that makes sense. I am installing it via `google-cloud`, but we have the `requests` library pinned to an earlier version, hence not pulling in `urllib3` automatically.

Pipfile would be useful here :-) Thanks for the insight.

Issue #42, opened 2017-10-24
Dependency on urllib3?


mlissner:
I just upgraded the packages on my system, and this package (which must be a dependency for another package) broke my system. I'm getting the error that:

```ImportError: No module named urllib3.response```

And I don't have urllib3 installed, so that's the problem. However, *why* don't I have it installed? I only use pip to install things, so I think the fact that urllib3 isn't installed means that it's not listed in a requirements.txt file somewhere.

It doesn't look like this project has a requirements list, but I don't know. It does seem like the urllib3 requirement might be a new one?

**Comment by mlissner** [comment]
Yeah, this looks like the culprit from three days ago: https://github.com/GoogleCloudPlatform/google-resumable-media-python/commit/b6c62d80ad8415626fade6caf4282939862996bd

I imagine this will be breaking a fair number of people's systems soon?

**Comment by dhermes** [comment]
@mlissner Can you be a bit more explicit about how this broke your system?

FWIW, `urllib3` comes in as a dependency of `requests`, which is an "extra" rather than being a direct dependency. In other words, a perfectly reasonable way to install this library would be

```
pip install google-resumable-media[requests]
```

/cc @jonparrott

**Comment by john2x** [comment]
We're currently affected by this issue. I was able to track it down to having an explicit dependency to the following:

```
requests==2.10.0
google-cloud==0.27
```

`requests 2.10.0` doesn't depend on `urllib3` yet, so it isn't installed automatically.

Updating our requirements to

```
requests==2.18.4
google-cloud==0.27
```

addresses the issue.

mlissner:
Yep, we have requests pinned with:

    requests==2.9.1

I think the fix here is to require a version of requests that contains urllib3 (or not use urllib3).

mlissner:
> Can you be a bit more explicit about how this broke your system?

And to respond to this....it broke my system because as part of deployment we always upgrade the packages in our requirements.txt file. In there we have requests pinned, but not `google-cloud`. So when we upgraded everything, google-cloud suddenly needed urllib3, but didn't have it.

dhermes:
> `google-cloud` suddenly needed `urllib3`

How do you mean? The dependency is optional because it can only be triggered if the `google.resumable_media.requests` package is imported. So you wouldn't see any issue at install time, just at run time. If you have code that is failing at run time, can you share a stacktrace? Also, can you share the pinned versions for `requests` and any package that begins with `google`?

----

[...]

PR #44, opened 2017-11-03
Making `requests` a strict dependency for the `requests` subpackage.

dhermes:
See #42 and #43.

/cc @brandonw @mlissner @tseaver

### DIFF
--- a/google/resumable_media/requests/__init__.py
+++ b/google/resumable_media/requests/__init__.py
@@ -659,6 +659,18 @@ transmitted in chunks until completion:
    True
 """
 
+import pkg_resources
+
+try:
+    pkg_resources.require('requests >= 2.18.0')
+except pkg_resources.ResolutionError as caught_exc:  # pragma: NO COVER
+    import six
+    new_exc = ImportError(
+        '``requests >= 2.18.0`` is required by the '
+        '``google.resumable_media.requests`` subpackage.\n'
+        'It can be installed via\n'
+        '    pip install google-resumable-media[requests].')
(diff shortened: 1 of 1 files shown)


google/resumable_media/requests/__init__.py
```diff
@@ -659,6 +659,18 @@ def SimpleUpload(*args, **kwargs):
    True
 """
 
+import pkg_resources
+
+try:
+    pkg_resources.require('requests >= 2.18.0')
```
dhermes: @jonparrott Should I make this less strict / less new?

I only need `requests.models.Response` to be a context manager and `urllib3` to be a separate dependency rather than vendored/bundled in.

dhermes:
I am going to merge this as-is, but we can continue the discussion @jonparrott. The docs failure is just because a [new version][1] of Sphinx has caused a changed in some of the "generated" docs files, so I'll resolve in a follow-up.

[1]: https://pypi.org/project/Sphinx/1.6.5/

google/resumable_media/requests/__init__.py
jonparrott: I don't feel strongly.

dhermes: Cool, we shall let it be then unless we hear otherwise.

dhermes:
I just pushed the docs commit to `master`: https://github.com/GoogleCloudPlatform/google-resumable-media-python/commit/50f4c4d22cdaea71c794639226e819197f11f555

PR #47, opened 2018-03-01
Update the default UPLOAD_CHUNK_SIZE to 100MB.

craigcitro:
The goal here is to lower latency for transfers, as well as having some
consistency between client libraries.

For more discussion, see
https://github.com/google/google-api-python-client/pull/482.

PTAL @dhermes -- I also didn't see an obvious place you were setting a download chunk size, but if you point me at one I'm happy to update it. 😁

### DIFF
--- a/google/resumable_media/common.py
+++ b/google/resumable_media/common.py
@@ -22,7 +22,7 @@ _SLEEP_RETRY_ERROR_MSG = (
     u'At most one of `max_cumulative_retry` and `max_retries` '
     u'can be specified.')
 
-UPLOAD_CHUNK_SIZE = 262144  # 256 * 1024
+UPLOAD_CHUNK_SIZE = 104857600  # 100 * 1024 * 1024
 """int: Chunks in a resumable upload must come in multiples of 256 KB."""
 PERMANENT_REDIRECT = 308
 """int: Permanent redirect status code.


dhermes:
RE: download chunk size, these are the two most relevant spots:

- [`ChunkedDownload.__init__`][1] does not set a default chunk size (i.e. the user must pass it)
- The `requests` implementation of [`Download._write_to_stream`][2] actual does use an "iterable response" with a chunk size of 8KB (i.e. one page)

[1]: https://github.com/GoogleCloudPlatform/google-resumable-media-python/blob/50f4c4d22cdaea71c794639226e819197f11f555/google/resumable_media/_download.py#L216
[2]: https://github.com/GoogleCloudPlatform/google-resumable-media-python/blob/50f4c4d22cdaea71c794639226e819197f11f555/google/resumable_media/requests/download.py#L122

dhermes:
This LGTM, but @jonparrott should LGTM too before a merge.

dhermes:
Eek, lots of "256 KB must divide chunk size" errors.

craigcitro:
So looking at the way that constant is used, I think I had misunderstood its purpose.

AFAICT, it's not used as the default for chunk sizes when starting an upload. Instead, the caller *always* provides `chunk_size` to the constructor, and the code is simply checking that the provided value is divisible by `UPLOAD_CHUNK_SIZE`. So any default that exists is probably in `google-cloud-storage` and/or other libraries that use this.

Is that correct? If so, I'll drop this PR, but my offer to tweak default chunksizes can be happily reapplied to `google-cloud-storage` if you'd like at no additional charge. 😀 

(The error message above was misleading, due to the string `256 KB` being hardcoded; sent #48 to fix.)

dhermes:
@craigcitro No that isn't correct. The actual storage API requires chunks are divisible by some value (I can't recall what it is).

craigcitro:
@dhermes We're maybe missing each other on this one. :)

Storage does indeed have a requirement on chunk size being divisible by 256KB: from the [docs](https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload), "Create chunks in multiples of 256 KB (256 x 1024 bytes) in size, except for the final chunk that completes the upload." (This same requirement holds for media transfers across all APIs.)

Now, in the `google-resumable-media` code, the user **always** [provides a chunk size](https://github.com/GoogleCloudPlatform/google-resumable-media-python/blob/8700f14e7c72501df557aee95c8a2426dc8c4f4c/google/resumable_media/_upload.py#L319) to the `ResumableUpload` constructor, and `UPLOAD_CHUNK_SIZE` is only used to confirm the invariant above for the provided value. 

My original goal in this CL was to change the **default** chunk size, but I think that was misguided: there is no default chunk size. If such a default exists, it's presumably in the callers of this library.

More directly:
* `UPLOAD_CHUNK_SIZE` in this library is playing the same role as [`CHUNK_SIZE_MULTIPLE`](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/34cedf3faec6c3d1d92d551939f8070993c5fa90/storage/google/cloud/storage/blob.py#L124) in `google.cloud.storage`
* This was inspired by google/google-api-python-client#481, where I changed `DEFAULT_CHUNK_SIZE`; there's no corresponding constant in this library. (Or if there is, I haven't found it.)

sduskis:
@craigcitro, this is an old PR.  Can we close it?

craigcitro:
Sure.

Issue #56, opened 2018-12-20
Systest Failure: 'test_chunked_download'.


crwilcox:
The following file fails the chunked_download test.

https://github.com/googleapis/google-resumable-media-python/blob/master/tests/system/requests/test_download.py#L63-L71

Test: tests/system/requests/test_download.py::test_chunked_download 

Possibly related to #49

crwilcox:
Tracking this internally 121376753

tseaver:
```python
____________________________ test_chunked_download _____________________________
Traceback (most recent call last):
  File "/home/tseaver/projects/agendaless/Google/src/grmp/tests/system/requests/test_download.py", line 411, in test_chunked_download
    total_bytes, actual_contents)
  File "/home/tseaver/projects/agendaless/Google/src/grmp/tests/system/requests/test_download.py", line 383, in consume_chunks
    response = download.consume_next_chunk(authorized_transport)
  File "/home/tseaver/projects/agendaless/Google/src/grmp/google/resumable_media/requests/download.py", line 225, in consume_next_chunk
    self._process_response(result)
  File "/home/tseaver/projects/agendaless/Google/src/grmp/google/resumable_media/_download.py", line 348, in _process_response
    callback=self._make_invalid)
  File "/home/tseaver/projects/agendaless/Google/src/grmp/google/resumable_media/_helpers.py", line 64, in header_required
    response, u'Response headers must contain header', name)
google.resumable_media.common.InvalidResponse: ('Response headers must contain header', 'content-length')
```

crwilcox:
Followup with some details:

If transfer-encoding is set, the spec dictates that "The Content-Length header field MUST NOT be sent if these two lengths are different (i.e., if a Transfer-Encoding header field is present)."
Because there's  Transfer-Encoding header set on that response. See RFC 2616 section 4.4, #3:

If a Content-Length header field (section 14.13) is present, its decimal value in OCTETs represents both the entity-length and the transfer-length. The Content-Length header field MUST NOT be sent if these two lengths are different (i.e., if a Transfer-Encoding header field is present).

tseaver:
I can tweak the handling for that, but then we run up against the same underlying issue in our system test for the gzipped file:  instead of getting the file in chunks, we get it all in one blast, because `requests` handles the unzipping for us.  We may need to investigate using [`response.raw`](http://2.python-requests.org/en/master/user/quickstart/?highlight=gzip#raw-response-content).

[...]

PR #74, opened 2019-07-23
GRMP: Fix system test chunk download and syntax error

HemangChothani:
Fixes: #56

### DIFF
--- a/tests/system/requests/test_download.py
+++ b/tests/system/requests/test_download.py
@@ -63,12 +63,12 @@ ALL_FILES = (
     }, {
         u'path': os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt.gz')),
         u'uncompressed':
-            os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt')),
+            os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt.gz')),
         u'content_type': PLAIN_TEXT,
         u'checksum': u'KHRs/+ZSrc/FuuR4qz/PZQ==',
         u'slices': (),
         u'metadata': {
-            u'contentEncoding': u'gzip',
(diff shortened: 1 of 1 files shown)


HemangChothani:
`System tests` are already failed in master branch as got the result `4 failed, 14 passed, 2 xfailed, 6 error in 101.97 seconds`  and for the  current branch `4 failed, 15 passed, 1 xfailed, 6 error in 107.77 seconds`.

tests/system/requests/test_download.py
```diff
@@ -63,19 +63,20 @@
     }, {
         u'path': os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt.gz')),
         u'uncompressed':
-            os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt')),
+            os.path.realpath(os.path.join(DATA_DIR, u'gzipped.txt.gz')),
```
tseaver: This change defeats the purpose of the `uncompressed` key, which is to point to the file in its un-gzipped form.

tests/system/requests/test_download.py
```diff
@@ -63,19 +63,20 @@
...
         u'content_type': PLAIN_TEXT,
         u'checksum': u'KHRs/+ZSrc/FuuR4qz/PZQ==',
         u'slices': (),
         u'metadata': {
-            u'contentEncoding': u'gzip',
+            u'Content-Encoding': u'gzip',
```
tseaver: This change is incorrect:  the key in the [Object representation](https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations) is `contentEncoding`.  By not passing it correctly, you hide the fact that the file is gzipped from the back-end, which then doesn't offer to serve it that way.  In conjunction with the change above, this change removes the whole point of testing with this file.

PR #78, opened 2019-07-25
Feature timeout in transport

### LINKED ISSUE #45: Add request timeout
Hi
it seems that our project, which use https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/storage/google/cloud/storage, is getting stuck during upload.

Although we can't confirm it with absolute certainty, it may be good idea have request timeout in resumable-media anyway.
Questions are:
- do you find a timeout good idea in general?
- should be there some well-chosen conservative implicit timeout by library (or explicit user defined timeout is preferred despite much complicated implementation because this library is only used by other google cloud platform libraries)?

We can provide eg. pull request for default timeout but we rather discuss all first.

pchauhan-qlogic:
fixes #45  and  #7

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -136,7 +136,7 @@ class Download(_helpers.RequestsMixin, _download.Download):
                 self.media_url, expected_md5_hash, actual_md5_hash)
             raise common.DataCorruption(response, msg)
 
-    def consume(self, transport):
+    def consume(self, transport, **transport_kwargs):
         """Consume the resource to be downloaded.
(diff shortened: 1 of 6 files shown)


crwilcox:
We are actively working on a design for retry/timeout but not ready to accept PRs for this at this time.

crwilcox:
@pchauhan-qlogic while we closed this, it may be useful to make some notes. Can you explain the TimeoutAuthorizedSession and its purpose. The rest of this PR is mostly about plumbing kwargs, but is there something more there?

PR #87, opened 2019-08-14
Always use raw response data.

### LINKED ISSUE #37: deflate-encoded content will fail md5 checksum
https://github.com/GoogleCloudPlatform/google-resumable-media-python/pull/36 fixes the problem of MD5 checksum failures when content was stored with content-encoding:gzip, but we will have a similar problem for content with content-encoding:deflate because that fix only monkey patches the decoder for the gzip case.

Two options:

1. Change accept-encoding to only be "gzip", rather than "gzip, deflate"
2. Broaden the monkey patch approach stared in pull/36 to support deflate as well.

(Side note: I've read that deflate is somewhat more CPU efficient than gzip, which I think is more of a concern for browsers in general.)

### LINKED ISSUE #49: `_write_to_stream()` in the `Download` class forces decompression of the response even if "Accept-Encoding: gzip" was present in the request

### LINKED ISSUE #50: md5 validation fails on gzipped PNG image
This is essentially the same issue as was fixed in #36, but now occuring specifically for PNG files with `Content-Type image/png` and `Content-Encoding gzip`. I can successfully download neighboring ascii files that are also gzipped (but without a `Content-Type` set).

This occurs with `google-resumable-media` 0.3.1:
```
> pip3 freeze | grep google
google-api-core==1.4.0
google-auth==1.5.1
google-cloud-core==0.28.1
google-cloud-storage==1.11.0
google-resumable-media==0.3.1
googleapis-common-protos==1.5.3
```

If I downgrade to `google-resumable-media` 0.2.3 the PNG download succeeds.

Error info:
```
/usr/local/lib/python3.6/dist-packages/google/cloud/storage/blob.py:538: in download_to_file                                                                                           
    transport, file_obj, download_url, headers, start, end)                                                                                                                            
/usr/local/lib/python3.6/dist-packages/google/cloud/storage/blob.py:480: in _do_download                                                                                               
    download.consume(transport)                                                                                                                                                        
/usr/local/lib/python3.6/dist-packages/google/resumable_media/requests/download.py:174: in consume
    self._write_to_stream(result)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <google.resumable_media.requests.download.Download object at 0x7f81c0f38cf8>, response = <Response [200]>

    def _write_to_stream(self, response):
        """Write response body to a write-able stream.
    
            .. note:
    
                This method assumes that the ``_stream`` attribute is set on the
                current download.
    
            Args:
                response (~requests.Response): The HTTP response object.
    
            Raises:
                ~google.resumable_media.common.DataCorruption: If the download's
                    checksum doesn't agree with server-computed checksum.
            """
        expected_md5_hash = self._get_expected_md5(response)                                                                                                                  [20/1803]
    
        if expected_md5_hash is None:
            md5_hash = _DoNothingHash()
        else:
            md5_hash = hashlib.md5()
        with response:
            # NOTE: This might "donate" ``md5_hash`` to the decoder and replace
            #       it with a ``_DoNothingHash``.
            local_hash = _add_decoder(response.raw, md5_hash)
            body_iter = response.iter_content(
                chunk_size=_SINGLE_GET_CHUNK_SIZE, decode_unicode=False)
            for chunk in body_iter:
                self._stream.write(chunk)
                local_hash.update(chunk)
    
        if expected_md5_hash is None:
            return
    
        actual_md5_hash = base64.b64encode(md5_hash.digest())
        # NOTE: ``b64encode`` returns ``bytes``, but ``expected_md5_hash``
        #       came from a header, so it will be ``str``.
        actual_md5_hash = actual_md5_hash.decode(u'utf-8')
        if actual_md5_hash != expected_md5_hash:
            msg = _CHECKSUM_MISMATCH.format(
                self.media_url, expected_md5_hash, actual_md5_hash)
>           raise common.DataCorruption(response, msg)
E           google.resumable_media.common.DataCorruption: Checksum mismatch while downloading:
E           
E             https://www.googleapis.com/download/storage/v1/b/****/o/******.png?alt=media
E           
E           The X-Goog-Hash header indicated an MD5 checksum of:
E           
E             rlofKEyAqWgDmEc3WO61BA==
E           
E           but the actual MD5 checksum of the downloaded contents was:
E           
E             h9fia1VRkKYNo656ukQ4hQ==

/usr/local/lib/python3.6/dist-packages/google/resumable_media/requests/download.py:137: DataCorruption
```

### LINKED ISSUE #56: Systest Failure: 'test_chunked_download'.

### LINKED ISSUE #76: Systest Failure: 'test_corrupt_download'.
```python
____________________________ test_corrupt_download _____________________________
Traceback (most recent call last):
  File "/home/tseaver/projects/agendaless/Google/src/grmp/tests/system/requests/test_download.py", line 239, in test_corrupt_download
    assert exc_info.value.args == (msg,)
AssertionError: assert ('Checksum mi...jgQ4nGw==\n',) == ('Checksum mis...4qz/PZQ==\n',)
  At index 0 diff: 'Checksum mismatch while downloading:\n\n  https://www.googleapis.com/download/storage/v1/b/grpm-systest-1563999706247/o/file.txt?alt=media\n\nThe X-Goog-Hash header indicated an MD5 checksum of:\n\n  1B2M2Y8AsgTpgAmY7PhCfg==\n\nbut the actual MD5 checksum of the downloaded contents was:\n\n  XHSHAr/SpIeZtZbjgQ4nGw==\n' != 'Checksum mismatch while downloading:\n\n  https://www.googleapis.com/download/storage/v1/b/grpm-systest-1563999706247/o/file.txt?alt=media\n\nThe X-Goog-Hash header indicated an MD5 checksum of:\n\n  1B2M2Y8AsgTpgAmY7PhCfg==\n\nbut ...
  
  ...Full output truncated (2 lines hidden), use '-vv' to show
```

tseaver:
Don't expect contra-spec `Content-Length` header for chunked downloads.

Re-enable Python 2.7 unit tests.

Closes #37
Closes #49
Closes #50
Closes #56
Closes #76

### DIFF
--- a/google/resumable_media/_download.py
+++ b/google/resumable_media/_download.py
@@ -343,18 +343,30 @@ class ChunkedDownload(DownloadBase):
         _helpers.require_status_code(
             response, _ACCEPTABLE_STATUS_CODES,
             self._get_status_code, callback=self._make_invalid)
-        content_length = _helpers.header_required(
-            response, u'content-length', self._get_headers,
-            callback=self._make_invalid)
-        num_bytes = int(content_length)
-        _, end_byte, total_bytes = get_range_info(
-            response, self._get_headers, callback=self._make_invalid)
+        headers = self._get_headers(response)
         response_body = self._get_body(response)
-        if len(response_body) != num_bytes:
-            self._make_invalid()
-            raise common.InvalidResponse(
-                response, u'Response is different size than content-length',
-                u'Expected', num_bytes, u'Received', len(response_body))
+
+        start_byte, end_byte, total_bytes = get_range_info(
+            response, self._get_headers, callback=self._make_invalid)
+
+        transfer_encoding = headers.get(u'transfer-encoding')
+
+        if transfer_encoding is None:
+            content_length = _helpers.header_required(
+                response, u'content-length', self._get_headers,
+                callback=self._make_invalid)
+            num_bytes = int(content_length)
+            if len(response_body) != num_bytes:
+                self._make_invalid()
+                raise common.InvalidResponse(
+                    response,
+                    u'Response is different size than content-length',
+                    u'Expected', num_bytes,
+                    u'Received', len(response_body),
+                )
+        else:
+            # 'content-length' header not allowed with chunked encoding.
+            num_bytes = end_byte - start_byte + 1
 
         # First update ``bytes_downloaded``.
         self._bytes_downloaded += num_bytes
--- a/google/resumable_media/requests/_helpers.py
+++ b/google/resumable_media/requests/_helpers.py
@@ -25,6 +25,7 @@ from google.resumable_media import common
 
 
 _DEFAULT_RETRY_STRATEGY = common.RetryStrategy()
+_SINGLE_GET_CHUNK_SIZE = 8192
 
 
 class RequestsMixin(object):
@@ -69,7 +70,12 @@ class RequestsMixin(object):
         Returns:
             bytes: The body of the ``response``.
         """
-        return response.content
+        if response._content is False:
+            response._content = b''.join(
+                response.raw.stream(
+                    _SINGLE_GET_CHUNK_SIZE, decode_content=False))
+            response._content_consumed = True
+        return response._content
 
 
 def http_request(transport, method, url, data=None, headers=None,
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -18,15 +18,12 @@ import base64
 import hashlib
 import logging
 
-import urllib3.response
-
 from google.resumable_media import _download
 from google.resumable_media import common
 from google.resumable_media.requests import _helpers
 
 
 _LOGGER = logging.getLogger(__name__)
-_SINGLE_GET_CHUNK_SIZE = 8192
 _HASH_HEADER = u'x-goog-hash'
 _MISSING_MD5 = u"""\
 No MD5 checksum was returned from the service while downloading {}
@@ -117,12 +114,12 @@ class Download(_helpers.RequestsMixin, _download.Download):
         with response:
             # NOTE: This might "donate" ``md5_hash`` to the decoder and replace
             #       it with a ``_DoNothingHash``.
-            local_hash = _add_decoder(response.raw, md5_hash)
-            body_iter = response.iter_content(
-                chunk_size=_SINGLE_GET_CHUNK_SIZE, decode_unicode=False)
+            body_iter = response.raw.stream(
+                _helpers._SINGLE_GET_CHUNK_SIZE, decode_content=False)
             for chunk in body_iter:
                 self._stream.write(chunk)
-                local_hash.update(chunk)
+                md5_hash.update(chunk)
+            response._content_consumed = True
 
         if expected_md5_hash is None:
             return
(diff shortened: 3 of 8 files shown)


Plantain:
This certainly broke my usecase, which is storing data in GCS gzipped (why wouldn't you?) and downloading it via python with download_as_string().
It took me a full day to debug that google-resumable-media 0.4.0 was responsible for this - basically no one consumes google-resumable-media directly so publishing it in your own release notes is not useful for users who depend on google-cloud-storage.
Reading the linked issues I'm not sure what real world issue was being fixed here to make it worthwhile to break every user in an obtuse way using a combination of download_as_string (https://googleapis.github.io/google-cloud-python/latest/_modules/google/cloud/storage/blob.html#Blob.download_as_string) and gzip.

crwilcox:
@Plantain I am currently working to revert this since it has implications for existing users of storage. We will bring this back but in a way that doesn't break your use. Sorry for any time you spent figuring this out.

PR #103, opened 2019-09-16
Revert "Always use raw response data. (#87)"

crwilcox:
This reverts commit 2b9ffc813b47912b5609985db9dc5d9e4936310a.

This library is being used by google-cloud-storage with only a >= specifier. This change caused breakages in all existing clients.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -24,7 +24,7 @@ SYSTEM_TEST_ENV_VARS = (
 GOOGLE_AUTH = 'google-auth >= 0.10.0'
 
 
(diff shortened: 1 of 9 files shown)


noxfile.py
```diff
@@ -24,7 +24,7 @@
 GOOGLE_AUTH = 'google-auth >= 0.10.0'
 
 
-@nox.session(python=['2.7', '3.4', '3.5', '3.6', '3.7'])
+@nox.session(python=['2,7', '3.4', '3.5', '3.6', '3.7'])
```
frankyn: Trying not to nit a revert but it might be okay to keep this fix.

Issue #106, opened 2019-09-17
Reimplement changes undone by #103 (always use raw response data)


crwilcox:
Changes originally merged in #87 were undone in #103 in a way that can be opted in to to avoid breaking usage by users on older client library versions. Particularly GCS

crwilcox:
This issue over on GCS, https://github.com/googleapis/google-cloud-python/issues/9188, seems to point to us making it more clear when we are decompressing and what the encoding is.

tseaver:
See PR #87.  Note that redoing them is complicated by the fact that the code has since been blackened. :(

tseaver:
Plan:

- [ ] Restore partial fix for #56. (PR #107).
- [ ] Split out decoding vs. raw-only behavior into separate classes.

PR #107, opened 2019-09-18
Don't expect 'Content-Length' header for chunked downloads.

tseaver:
Toward #106.

Restores partial fix for #56, reverted by #103.


tseaver:
Rather than review this separately, I'll roll it up in the PR which re-adds optional raw downloads.

PR #109, opened 2019-09-26
feat: add raw download classes

### LINKED ISSUE #106: Reimplement changes undone by #103 (always use raw response data)

tseaver:
Restores "raw download" behavior as separate `RawDownload` and `RawChunkedDownload` classes:  the existing `Download` and `ChunkedDownload` classes retain their bad-but-backward-compatible behavior of expanding resources with `Content-Encoding` headers set.

Closes #106.

### DIFF
--- a/tests/system/requests/test_download.py
+++ b/tests/system/requests/test_download.py
@@ -40,9 +40,40 @@ NOT_FOUND_ERR = (
 )
 
 
+class CorruptingAuthorizedSession(tr_requests.AuthorizedSession):
+    """A Requests Session class with credentials, which corrupts responses.
+
+    This class is used for testing checksum validation.
+
+    Args:
+        credentials (google.auth.credentials.Credentials): The credentials to
+            add to the request.
+        refresh_status_codes (Sequence[int]): Which HTTP status codes indicate
+            that credentials should be refreshed and the request should be
+            retried.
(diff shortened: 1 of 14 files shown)


tseaver:
I have verified that the `google-cloud-storage` system tests pass with this branch installed:  of course, none of them take advantage (yet) of the new raw download classes.

tseaver:
@crwilcox PTAL

tests/system/requests/test_download.py
```diff
@@ -58,60 +98,42 @@
         ),
     },
     {
-        u"path": os.path.realpath(os.path.join(DATA_DIR, u"file.txt")),
+        u"path": get_path(u"file.txt"),
         u"content_type": PLAIN_TEXT,
-        u"checksum": u"KHRs/+ZSrc/FuuR4qz/PZQ==",
+        u"checksum": u"XHSHAr/SpIeZtZbjgQ4nGw==",
```
crwilcox: just verifying that this isn't a breaking change?

tseaver: Yup.  We are no longer validating checksums for non-raw downloads, so the old "expanded" checksum is not valid.

PR #114, opened 2019-11-13
feat: add 'response_callback' to 'consume' / 'consume_next_chunk'

### LINKED ISSUE #113: Add 'response_callback' argument to 'consume' / 'consume_next_chunk' methods.
In support of https://github.com/googleapis/google-cloud-python/issues/9787.

Whether or not the implementation in `google-cloud-storage` exposes a callback argument, we need to provide it here, to allow either `Blob.download_*` or its callers access to the response headers.

tseaver:
Allows callers access to the response objects, e.g., to use HTTP headers.

Closes #113.


tseaver:
@william-silversmith has submitted a [PR to `python-storage`](https://github.com/googleapis/python-storage/pull/204) which makes the general-purpose callback strategy unneeded.

PR #116, opened 2020-01-08
feat(resumable-media): add customizable timeouts to upload/download methods

### LINKED ISSUE #45: Add request timeout

plamut:
Fixes #45.

This PR adds configurable timeouts to `upload` objects.

**Motivation:**
We would like to add timeouts to BigQuery methods that depend on the file upload functionality (using resumable media), so that requests do not get stuck indefinitely at the transport layer.

Currently there is a default timeout used (61, 60), but that might not suit all use cases, hence the need for optional explicit timeout parameters.

> **CAUTION:** Changing public function signatures might break some libraries that utilize resumable media, even though the new arguments are optional, and positioned last), and this will not be caught by the test suite.
>
> It has happened recently when we added timeouts to `google-auth`, thus it is advisable that code owners of the dependent libraries have a look, too (if that is feasible).

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,0 +1,1 @@
+releaseType: python
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 # Copyright 2018 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/.kokoro/publish-docs.sh
+++ b/.kokoro/publish-docs.sh
@@ -1,4 +1,17 @@
 #!/bin/bash
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 set -eo pipefail
 
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -1,4 +1,17 @@
 #!/bin/bash
+# Copyright 2020 Google LLC
+#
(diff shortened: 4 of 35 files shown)


plamut:
Unit tests suite fails under Python 3.4, created an issue (https://github.com/googleapis/google-resumable-media-python/issues/117). The 3.4 tests pass, though, if ran in isolation.

anantzoid:
I've been facing `ReadTimeout` issues while reading the data from GCS and noticed `_DEFAULT_READ_TIMEOUT = 60` cannot be changed by the user for read requests [[src](https://github.com/googleapis/google-resumable-media-python/blob/master/google/resumable_media/requests/_helpers.py#L34)].

The user should be able to specify timeout while using [`download_to_file`](https://googleapis.dev/python/storage/latest/blobs.html).

plamut:
@busunkim96 Somehow this slipped out of the focus, and it might be a good time to revive it. Added you as a reviewer based on the GitHub's suggestion, but feel free to delegate to someone else if that makes more sense, thanks!

@anantzoid Thanks for bringing this back to attention. As this PR only adds configurable timeouts to uploads, I can add the same for the download methods, too, provided that the original [feature request](https://github.com/googleapis/google-resumable-media-python/issues/45) is confirmed (I personally don't see any reason why it shouldn't be).

busunkim96:
LGTM'd, but adding @crwilcox since the storage folks are other primary users of this library

plamut:
I will update the PR with timeouts to the download methods as well.

plamut:
@crwilcox Friendly ping, can you take a look? Thanks!

plamut:
@busunkim96 @frankyn Thanks for the reviews!

When would it be possible to release a new version of `google-resumable-media` in order to unblock https://github.com/googleapis/python-storage/pull/185 ? I hear this week is a holiday period there?

frankyn:
Hi @plamut, let's wait until next week when everyone is back.

plamut:
Sounds good, there's no rush.

PR #135, opened 2020-06-19
feat: add configurable crc32c checksumming for downloads

andrewsg:


andrewsg:
Received a design suggestion to move the argument to the class constructor for the download class; looking into refactoring that now.

google/resumable_media/requests/download.py
```diff
@@ -82,44 +82,67 @@ def _write_to_stream(self, response):
...
+            if actual_checksum != expected_checksum:
+                msg = _CHECKSUM_MISMATCH.format(
+                    self.media_url,
+                    expected_checksum,
+                    actual_checksum,
+                    checksum_type=checksum.upper(),
+                )
+                raise common.DataCorruption(response, msg)
+
+    def consume(self, transport, checksum="md5"):
```
frankyn, line 145: I think you need to still update this bit to be populated at Init adding a comment to make sure I don't miss in subsequent reviews.

google/resumable_media/requests/download.py
```diff
@@ -456,21 +520,21 @@ def _parse_md5_header(header_value, response):
 class _DoNothingHash(object):
     """Do-nothing hash object.
 
-    Intended as a stand-in for ``hashlib.md5`` in cases where it
-    isn't necessary to compute the hash.
+    Intended as a stand-in for ``hashlib.md5`` or another checksum in cases
```
frankyn, line 523: Suggestion:

`For md5 and crc32c when it isn't necessary to compute the hash.`

google/resumable_media/requests/download.py
```diff
@@ -194,44 +224,64 @@ def _write_to_stream(self, response):
...
+            #       came from a header, so it will be ``str``.
+            actual_checksum = actual_checksum.decode(u"utf-8")
+            if actual_checksum != expected_checksum:
+                msg = _CHECKSUM_MISMATCH.format(
+                    self.media_url,
+                    expected_checksum,
+                    actual_checksum,
+                    checksum_type=checksum.upper(),
+                )
+                raise common.DataCorruption(response, msg)
```
frankyn, line 282: Do you know if data will still be successfully downloaded then raise this DataCorruption() error?

andrewsg: I haven't changed this behavior from the previous iteration but my understanding is that actually deleting a download is the responsibility of the user, for instance because we support writing to streams or other situations where it could be unsafe to delete or rewind without user intervention.

google/resumable_media/requests/download.py
andrewsg: I'd prefer to be specific about the classes it's emulating so I'll update this to be specific both about this class and that crc32c is accepted.

andrewsg:
@frankyn PTAL

google/resumable_media/_download.py
```diff
@@ -124,8 +124,23 @@ class Download(DownloadBase):
             ``start`` to the end of the media.
         headers (Optional[Mapping[str, str]]): Extra headers that should
             be sent with the request, e.g. headers for encrypted data.
+        checksum Optional([str]: The type of checksum to compute to verify
+            the integrity of the object. The response headers must contain
+            a checksum of the requested type. If the headers lack an
+            appropriate checksum (for instance in the case of transcoded or
+            ranged downloads where the remote service does not know the
+            correct checksum) an INFO-level log will be emitted. Supported
+            values are "md5", "crc32c" and None.
```
frankyn, line 133: Could you call out the default value in here as well.

google/resumable_media/_download.py
```diff
@@ -124,8 +124,23 @@ class Download(DownloadBase):
...
     """
 
+    def __init__(
+        self, media_url, stream=None, start=None, end=None, headers=None, checksum="md5"
```
frankyn, line 137: @crwilcox per discussion with @andrewsg, the only gotcha with checksum being set in the constructor is that a user can manipulate the value after initialization and cause a failure during consume. 

I think this may an edge case and not too concerned. Bringing it up as remaining talking point.

andrewsg: I addressed this by doing validation in the consume method, not in the constructor. Manipulating the value after initialization will not cause an issue.

frankyn: Sounds good, then that just leaves my other comment on documentation. 

Thanks @andrewsg

andrewsg:
Fixed merge conflict and addressed comments.

google/resumable_media/_download.py
```diff
@@ -124,8 +124,23 @@ class Download(DownloadBase):
             ``start`` to the end of the media.
         headers (Optional[Mapping[str, str]]): Extra headers that should
             be sent with the request, e.g. headers for encrypted data.
+        checksum Optional([str]: The type of checksum to compute to verify
```
frankyn, line 127: missing `)` after `str]`

google/resumable_media/requests/download.py
```diff
@@ -87,37 +94,53 @@ def _write_to_stream(self, response):
...
-        expected_md5_hash = _get_expected_md5(
-            response, self._get_headers, self.media_url
-        )
+        if self.checksum not in ["md5", "crc32c", None]:
+            raise ValueError("checksum must be ``'md5'``, ``'crc32c'`` or ``None``")
+
+        if self.checksum:
+            # `_get_expected_checksum()` may return None, in which case it will
+            # emit the info log `_MISSING_CHECKSUM`.
+            expected_checksum = _get_expected_checksum(
```
frankyn, line 103: Missed a potential refactor of code. This is a suggestion based on conditions being checked.

Could you make `_get_expected_checksum()` return checksum_object as well?

IIUC there's a check on checksum_type so you could remove the block below:

```python
        if expected_checksum is not None and self.checksum == "md5":
            checksum_object = hashlib.md5()
        elif expected_checksum is not None and self.checksum == "crc32c":
            checksum_object = _helpers._get_crc32c_object()
        else:
            md5_hash = hashlib.md5()
            checksum_object = _DoNothingHash()
```

andrewsg:
Thanks for the suggestion, I was able to remove some duplication of code between Download and RawDownload and tighten it up a bit in general. PTAL

google/resumable_media/_download.py
frankyn: Still missing ending `)`.

google/resumable_media/requests/_helpers.py
```diff
@@ -33,6 +34,12 @@
 # The number of seconds to wait between bytes sent from the server.
 _DEFAULT_READ_TIMEOUT = 60
 
+_SLOW_CRC32C_WARNING = (
+    "Currently using crcmod in pure python form. This is a slow "
+    "implementation. You may consider installing `google-crc32c` for a "
+    "precompiled wheel with fast crc32c support."
+)
```
frankyn, line 41: IIUC this is only possible when Python is 2.x. I would get confused as a user because google-crc32c python package is a requirement for installation.

andrewsg: Okay, PTAL

andrewsg:
@frankyn fixed coverage issue; PTAL

PR #170, opened 2020-09-08
fix: allow space between checksums in header

### LINKED ISSUE #169: Allow space in-between checksums in response headers
**Summary:** I encountered this bug when downloading files using the Google Cloud Storage Python libraries, but the bug is coming from the Resumable Media library.  The current code (and [official docs](https://cloud.google.com/storage/docs/xml-api/reference-headers#xgooghash)) assumes the checksum headers will be contiguous without whitespace:
```
'x-goog-hash': 'crc32c=t6zgHw==,md5=zH3E9XwJPGGra4vnT+aamQ=='
```
However, I'm actually getting responses with whitespaces between each type of checksum:
```
'x-goog-hash': 'crc32c=t6zgHw==, md5=zH3E9XwJPGGra4vnT+aamQ=='
```

Because of this, the checksum of downloaded files aren't being verified.

Solution is to patch `_helpers.py` as follows:
```
def _parse_checksum_header(header_value, response, checksum_label):
    if header_value is None:
        return None

    matches = []
    for checksum in header_value.split(u","):
        name, value = checksum.split(u"=", 1)
        # Use lstrip to support both "," and ", " as delimiters
        if name.lstrip() == checksum_label:
            matches.append(value)

    if len(matches) == 0:
        return None
    elif len(matches) == 1:
        return matches[0]
    else:
        raise common.InvalidResponse(
            response,
            u"X-Goog-Hash header had multiple ``{}`` values.".format(checksum_label),
            header_value,
            matches,
        )
```

Relevant: #22, #31, #32

mcsimps2:
The current checksum logic only allowed checksum headers in the format `crc32c=hash1,md5=hash2`, but GCP may actually return them with a space in-between as `crc32c=hash1, md5=hash2`.

This means that checksums are not being verified correctly for any header response like the latter.

If you want an actual example of a response from GCP that demonstrates this, here's one:
```
{
   "X-GUploader-UploadID":"REDACTED",
   "Expires":"Tue, 08 Sep 2020 21:27:47 GMT",
   "Date":"Tue, 08 Sep 2020 21:27:47 GMT",
   "Cache-Control":"private, max-age=0",
   "Last-Modified":"Tue, 08 Sep 2020 21:27:47 GMT",
   "ETag":"REDACTED",
   "x-goog-generation": "REDACTED",
   "x-goog-metageneration":"1",
   "x-goog-stored-content-encoding":"identity",
   "x-goog-stored-content-length":"1043237",
   "Content-Type":"application/octet-stream",
   "x-goog-hash":"crc32c=t6zgHw==, md5=zH3E9XwJPGGra4vnT+aamQ==",
   "x-goog-storage-class":"STANDARD",
   "Accept-Ranges":"bytes",
   "Content-Length":"1043237",
   "Server":"UploadServer"
}
```

Resolves #169

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -313,7 +313,8 @@ def _parse_checksum_header(header_value, response, checksum_label):
     matches = []
     for checksum in header_value.split(u","):
         name, value = checksum.split(u"=", 1)
-        if name == checksum_label:
+        # Official docs say "," is the separator, but real-world responses have encountered ", "
+        if name.lstrip() == checksum_label:
             matches.append(value)
 
     if len(matches) == 0:
--- a/tests/unit/test__helpers.py
+++ b/tests/unit/test__helpers.py
@@ -311,11 +311,12 @@ def test__DoNothingHash():
 
 
 class Test__get_expected_checksum(object):
+    @pytest.mark.parametrize("template", [u"crc32c={},md5={}", u"crc32c={}, md5={}"])
     @pytest.mark.parametrize("checksum", ["md5", "crc32c"])
     @mock.patch("google.resumable_media._helpers._LOGGER")
-    def test__w_header_present(self, _LOGGER, checksum):
+    def test__w_header_present(self, _LOGGER, template, checksum):
         checksums = {"md5": u"b2twdXNodGhpc2J1dHRvbg==", "crc32c": u"3q2+7w=="}
-        header_value = u"crc32c={},md5={}".format(checksums["crc32c"], checksums["md5"])
+        header_value = template.format(checksums["crc32c"], checksums["md5"])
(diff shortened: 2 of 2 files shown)


crwilcox:
Hi @mcsimps2 thank you for the contribution. Could I bother you to make a corresponding issue for this?

mcsimps2:
@crwilcox It's mentioned in the summary at the bottom - it resolves #169.  I should've made it more clear that was the issue of concern, sorry about that!

crwilcox:
@mcsimps2 Cool, I just missed it :)

crwilcox (approved):

crwilcox:
Tests are reporting failed for linting, which is not your pr. I can create a separate PR and clean that up.

PR #178, opened 2020-09-29
fix: move `async_resumable_media` to `_async_resumable_media`

crwilcox:
moves the async surface to internal while we allow early access. googleapis/google-cloud-python#15626 tracks moving it back once we are satisfied with quality, and this is actively used by a client.


tritone (commented):
Looks good. Maybe we should have a README at the top level of the async package noting that this is internal-only, as well? Not sure what the convention is in the python libraries.

crwilcox:
So, the naming (leading `_`) should be enough, but I can certainly add a blurb as a note. sure :)

tritone (approved):

PR #153, opened 2020-08-03
feat: async changes to resumable upload/download

anibadde:
AsyncIO functionality for resumable media

Current State: All unit tests for asynchronous functionality are passing. The system tests are close to complete, though a timeout context manager error means that we are currently not able to run the upload and download tests simultaneously and need to run each in separate sessions to avoid that bug. 

Related Branch for auth: https://github.com/googleapis/google-auth-library-python/tree/async

PR #179, opened 2020-09-29
feat: add _async_resumable_media experimental support

crwilcox:
This is a merge of multiple previously reviewed PRs
googleapis/google-resumable-media-python#178
googleapis/google-resumable-media-python#176
googleapis/google-resumable-media-python#153

It says 9k lines, keep in mind, technically this is all reviewed code. I think areas to focus on are changes to existing public interfaces, as the async code should all be boxed in an internal area. This is setup.py and noxfile.py


google/_async_resumable_media/requests/_request_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+This utilities are explicitly catered to ``requests``-like transports.
+"""
+
+
+import functools
+
+from google._async_resumable_media import _helpers
+from google.resumable_media import common
+
+import google.auth.transport._aiohttp_requests as aiohttp_requests
```
crwilcox, line 26: This, which is used to allow for a combined response, has a hard dependency on google-auth. We don't currently have one for the package.

google/_async_resumable_media/requests/_request_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+import aiohttp
```
crwilcox, line 27: requests is an optional extra, but due to the use of an aiohttp specific timeout, we have to make an explicit dependency.

google/_async_resumable_media/requests/_request_helpers.py
busunkim96: Should `google-auth` be re-added as an explicit dependency?

busunkim96:
`noxfile.py` looks good to me, just curious about the dependencies needed in `setup.py`.

google/_async_resumable_media/requests/_request_helpers.py
crwilcox: I think it is safe to do so. It is unlikely this library is used without google-auth or something else that requires it (api-core for instance).

@frankyn @tswast any opinions?

crwilcox: https://github.com/googleapis/google-resumable-media-python/issues/180 I made thinking we might need this.

I sort of think refactoring to avoid the dependency may be futile since folks will have the dependency anyway?

busunkim96: @crwilcox I agree - if most folks are going to have to install `google-auth` I'd prefer to have it in the requirements.

google/_async_resumable_media/requests/_request_helpers.py
crwilcox: Given the conversation on https://github.com/googleapis/google-auth-library-python/issues/618 I think we should keep aiohttp an extra for the immediate term, since we aren't exposing users to it.

andrewsg: Agree

tests_async/system/credentials.json.enc
```diff
@@ -0,0 +1,52 @@
+U2FsdGVkX1+wqu1+eVu6OPbPoE0lzIp3B11p8Rdbha1ukxXcsskegJdBjcUqQOav
```
andrewsg, line 1: We could probably do without duplicating these files, instead importing them from the main tests directory, but it's up to you.

andrewsg (approved):
Due to the length of the PR and the various thorough previous incremental reviews I had to skim this one. Thanks for all your hard work on this.

tests_async/system/credentials.json.enc
crwilcox: I am going to make a process issue to cover this work. I don't want to stretch this PR out if possible :)

crwilcox: https://github.com/googleapis/google-resumable-media-python/issues/181

PR #186, opened 2020-11-16
fix: add retries for 'requests.ConnectionError'

andrewsg:
Restructures retry strategy to check both status codes and exceptions; then, adds requests.ConnectionError as a retriable exception. (Note: the unified standard library ConnectionError class is not used because it is unsupported by both Requests and Python 2).

### DIFF
--- a/google/_async_resumable_media/_helpers.py
+++ b/google/_async_resumable_media/_helpers.py
@@ -19,9 +19,6 @@ import random
 import time
 
 
-from six.moves import http_client
-
-
 from google.resumable_media import common
 
 
--- a/google/_async_resumable_media/_helpers.py
+++ b/google/_async_resumable_media/_helpers.py
@@ -27,13 +27,6 @@ from google.resumable_media import common
 
 RANGE_HEADER = u"range"
 CONTENT_RANGE_HEADER = u"content-range"
-RETRYABLE = (  # ConnectionError is also retried, but is not listed here.
-    common.TOO_MANY_REQUESTS,  # 429
-    http_client.INTERNAL_SERVER_ERROR,  # 500
-    http_client.BAD_GATEWAY,  # 502
-    http_client.SERVICE_UNAVAILABLE,  # 503
-    http_client.GATEWAY_TIMEOUT,  # 504
-)
 
 _SLOW_CRC32C_WARNING = (
     "Currently using crcmod in pure python form. This is a slow "
@@ -173,7 +166,7 @@ async def wait_and_retry(func, get_status_code, retry_strategy):
         except ConnectionError as e:
             error = e
         else:
-            if get_status_code(response) not in RETRYABLE:
+            if get_status_code(response) not in common.RETRYABLE:
                 return response
 
         if not retry_strategy.retry_allowed(total_sleep, num_retries):
--- a/google/_async_resumable_media/_helpers.py
+++ b/google/_async_resumable_media/_helpers.py
@@ -27,12 +27,12 @@ from google.resumable_media import common
 
 RANGE_HEADER = u"range"
 CONTENT_RANGE_HEADER = u"content-range"
-RETRYABLE = (
-    common.TOO_MANY_REQUESTS,
-    http_client.INTERNAL_SERVER_ERROR,
-    http_client.BAD_GATEWAY,
-    http_client.SERVICE_UNAVAILABLE,
-    http_client.GATEWAY_TIMEOUT,
+RETRYABLE = (  # ConnectionError is also retried, but is not listed here.
+    common.TOO_MANY_REQUESTS,  # 429
+    http_client.INTERNAL_SERVER_ERROR,  # 500
+    http_client.BAD_GATEWAY,  # 502
+    http_client.SERVICE_UNAVAILABLE,  # 503
+    http_client.GATEWAY_TIMEOUT,  # 504
 )
 
 _SLOW_CRC32C_WARNING = (
@@ -162,24 +162,33 @@ async def wait_and_retry(func, get_status_code, retry_strategy):
         object: The return value of ``func``.
     """
 
-    response = await func()
-
-    if get_status_code(response) not in RETRYABLE:
-        return response
-
     total_sleep = 0.0
     num_retries = 0
     base_wait = 0.5  # When doubled will give 1.0
-    while retry_strategy.retry_allowed(total_sleep, num_retries):
+
+    while True:  # return on success or when retries exhausted.
+        error = None
+        try:
+            response = await func()
+        except ConnectionError as e:
+            error = e
+        else:
+            if get_status_code(response) not in RETRYABLE:
+                return response
+
+        if not retry_strategy.retry_allowed(total_sleep, num_retries):
+            # Retries are exhausted and no acceptable response was received. Raise the
+            # retriable_error or return the unacceptable response.
+            if error:
+                raise error
+
+            return response
+
         base_wait, wait_time = calculate_retry_wait(base_wait, retry_strategy.max_sleep)
+
         num_retries += 1
         total_sleep += wait_time
         time.sleep(wait_time)
-        response = await func()
-        if get_status_code(response) not in RETRYABLE:
-            return response
-
-    return response
 
 
 class _DoNothingHash(object):
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -28,12 +28,12 @@ from google.resumable_media import common
 
 RANGE_HEADER = u"range"
 CONTENT_RANGE_HEADER = u"content-range"
-RETRYABLE = (
-    common.TOO_MANY_REQUESTS,
-    http_client.INTERNAL_SERVER_ERROR,
-    http_client.BAD_GATEWAY,
-    http_client.SERVICE_UNAVAILABLE,
-    http_client.GATEWAY_TIMEOUT,
+RETRYABLE = (  # ConnectionError is also retried, but is not listed here.
+    common.TOO_MANY_REQUESTS,  # 429
+    http_client.INTERNAL_SERVER_ERROR,  # 500
+    http_client.BAD_GATEWAY,  # 502
+    http_client.SERVICE_UNAVAILABLE,  # 503
+    http_client.GATEWAY_TIMEOUT,  # 504
 )
 
 _SLOW_CRC32C_WARNING = (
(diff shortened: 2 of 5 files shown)


google/resumable_media/_helpers.py
```diff
@@ -162,23 +162,43 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
     base_wait = 0.5  # When doubled will give 1.0
-    while retry_strategy.retry_allowed(total_sleep, num_retries):
+
+    # Set the retriable_exception_type if possible. We expect requests to be
+    # present here and the transport to be using requests.exceptions errors,
+    # but due to loose coupling with the transport layer we can't guarantee it.
+    try:
+        import requests
+
+        retriable_exception_type = requests.exceptions.ConnectionError
```
tritone, line 175: I would name this something specific to connection error.

google/resumable_media/_helpers.py
```diff
@@ -162,23 +162,43 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
+
+            return response
+
         base_wait, wait_time = calculate_retry_wait(base_wait, retry_strategy.max_sleep)
+
         num_retries += 1
         total_sleep += wait_time
         time.sleep(wait_time)
-        response = func()
-        if get_status_code(response) not in RETRYABLE:
```
tritone, line 178: It's interesting that this now happens before the wait rather than after. Seems like a correct change though?

google/resumable_media/_helpers.py
```diff
@@ -28,12 +28,12 @@
 
 RANGE_HEADER = u"range"
 CONTENT_RANGE_HEADER = u"content-range"
-RETRYABLE = (
-    common.TOO_MANY_REQUESTS,
-    http_client.INTERNAL_SERVER_ERROR,
-    http_client.BAD_GATEWAY,
-    http_client.SERVICE_UNAVAILABLE,
-    http_client.GATEWAY_TIMEOUT,
+RETRYABLE = (  # requests.ConnectionError is also retried, but is not listed here.
```
tritone, line 31: Maybe note why this is the case (because it's an exception and not a status code).

google/resumable_media/_helpers.py
```diff
@@ -162,23 +162,43 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
     total_sleep = 0.0
     num_retries = 0
     base_wait = 0.5  # When doubled will give 1.0
-    while retry_strategy.retry_allowed(total_sleep, num_retries):
+
+    # Set the retriable_exception_type if possible. We expect requests to be
+    # present here and the transport to be using requests.exceptions errors,
+    # but due to loose coupling with the transport layer we can't guarantee it.
+    try:
+        import requests
```
tseaver, line 173: This module is "above" the transport-specific `requests` subpackage, which breaks the "alternate transports are possible" design goal.

Because `google.resumable_media._helpers.wait_and_retry` is only called from one place, `google.resumable_media.requests._request_helpers.http_request`, it seems like we could just move `wait_and_retry` into that module, which would remove the need for the conditional import, as well as preserving the transport-independence of `google.resumable_media._helpers`.

If so, we should also move `google._async_resumable_media._helpers.wait_and_retry` down into `google._async_resumable_media.requests._request_helpers` as well, to maintain the (really painfully duplicated) structure.

google/_async_resumable_media/_helpers.py
```diff
@@ -27,12 +27,12 @@
 
 RANGE_HEADER = u"range"
 CONTENT_RANGE_HEADER = u"content-range"
-RETRYABLE = (
-    common.TOO_MANY_REQUESTS,
-    http_client.INTERNAL_SERVER_ERROR,
-    http_client.BAD_GATEWAY,
-    http_client.SERVICE_UNAVAILABLE,
-    http_client.GATEWAY_TIMEOUT,
+RETRYABLE = (  # ConnectionError is also retried, but is not listed here.
```
tritone, line 30: Can this whole set of errors be factored out into a common location for sync/async?

google/resumable_media/_helpers.py
```diff
@@ -162,23 +162,43 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
+        retriable_exception_type = None
+
+    while True:  # return on success or when retries exhausted.
+        error = None
+        try:
+            response = func()
+        except retriable_exception_type as e:
+            error = e
+        else:
+            if get_status_code(response) not in RETRYABLE:
```
tritone, line 186: Note that this covers 2xx responses as well as non-retryable errors.

tseaver (dismissed):

tests/unit/test__helpers.py
```diff
@@ -231,6 +262,33 @@ def test_retry_exceeds_max_cumulative(self, randint_mock, sleep_mock):
...
+            _helpers.wait_and_retry(func, _get_status_code, retry_strategy)
+
+        assert func.call_count == 8
+        assert func.mock_calls == [mock.call()] * 8
+
+        assert randint_mock.call_count == 7
+        assert randint_mock.mock_calls == [mock.call(0, 1000)] * 7
+
+        assert sleep_mock.call_count == 7
+        sleep_mock.assert_any_call(1.875)
```
tritone, line 284: Where do these numbers come from?

google/resumable_media/_helpers.py
tseaver, line 175: The `unit_2-2.7` and `system-2.7` tests are failing here, with 

```python
  File "/tmpfs/src/github/google-resumable-media-python/google/resumable_media/_helpers.py", line 175, in wait_and_retry
    retriable_exception_type = requests.exceptions.ConnectionError
AttributeError: 'module' object has no attribute 'exceptions'
```

Importing `requests.exceptions` explicitly would probably fix that.

google/resumable_media/_helpers.py
andrewsg: The reason here is simply that the original code called func() both before the while loop (zeroth retry) and inside it (first retry and beyond). The new one only has one call to func(), inside the loop. Any retry strategy stuff only fires up after the first call is complete in both designs, but this one unifies the func() calls.

tests/unit/test__helpers.py
andrewsg: This is just copied from the above functional test, but the numbers are the expected exponential backoff times plus the randint_mock.side_effect times above.

google/resumable_media/_helpers.py
andrewsg: The try/except import model here will guarantee alternative transports work, though since ConnectionError is specific to the transport, connection errors can't be caught on alternative transports. Given that the only thing changing is awareness of connection errors, I'm not sure I agree there's value to moving this into http_request, unless http_request is where it should have been all along. All processing of HTTP response codes is in this file and not in http_request so it makes sense to keep it here.

google/resumable_media/_helpers.py
andrewsg: Will do.

google/resumable_media/_helpers.py
andrewsg: Will do.

google/_async_resumable_media/_helpers.py
andrewsg: Should be possible, yes. I'll try.

google/resumable_media/_helpers.py
andrewsg: Whoops, thanks for catching that. I should be consistent about which of requests.ConnectionError and requests.exceptions.ConnectionError I use here (they refer to the same exception) to avoid these issues. I'll standardize on one or another and change imports to reflect.

google/resumable_media/_helpers.py
andrewsg: The requests import is nominally temporary as Python 3 standardizes on ConnectionError (stdlib), but no idea when that will really be the standard in practice as requests does not support it yet.

andyhky:
Is it possible that this would also retry requests that have `ChunkedEncodingError`?

Seeing this in google-resumable-media==1.10 (python 2.7, but I imagine other versions are impacted):

```  File "/usr/local/lib/python2.7/dist-packages/google/cloud/storage/blob.py", line 900, in _do_download
    response = download.consume(transport, timeout=timeout)
  File "/usr/local/lib/python2.7/dist-packages/google/resumable_media/requests/download.py", line 171, in consume
    self._write_to_stream(result)
  File "/usr/local/lib/python2.7/dist-packages/google/resumable_media/requests/download.py", line 105, in _write_to_stream
    for chunk in body_iter:
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 756, in generate
    raise ChunkedEncodingError(e)
ChunkedEncodingError: ("Connection broken: error(104, 'Connection reset by peer')", error(104, 'Connection reset by peer'))```

**Comment by andrewsg** [comment]
@tritone @tseaver Coverage at 100% now, PTAL. Also I'd like your opinion on @andyhky's suggestion above regarding ChunkedEncodingError. My feeling is we should treat it as a ConnectionError here, do you agree? If so I'll add it to our supported exceptions.

File: google/resumable_media/_helpers.py [hunk 8]
```diff
@@ -349,6 +362,16 @@ def _get_checksum_object(checksum_type):
         raise ValueError("checksum must be ``'md5'``, ``'crc32c'`` or ``None``")
 
 
+def _get_connection_error_class():
+    """Get the exception error class.
+
+    This is a separate function for testing purposes."""
```
frankyn, line 368: Clarify "testing purposes" with detecting if requests is being used.

google/resumable_media/_helpers.py
```diff
@@ -162,23 +155,43 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
-        return response
-
     total_sleep = 0.0
     num_retries = 0
     base_wait = 0.5  # When doubled will give 1.0
-    while retry_strategy.retry_allowed(total_sleep, num_retries):
+
+    # Set the retriable_exception_type if possible. We expect requests to be
+    # present here and the transport to be using requests.exceptions errors,
+    # but due to loose coupling with the transport layer we can't guarantee it.
```
frankyn, line 164: Can you file an issue to track potentially improving this in the future? E.g. Reliance on having a specific transport requests.

frankyn (changes requested):
Overall LGTM, just have nits on commentary.

google/resumable_media/_helpers.py
tritone: +1, good to explain this layer of indirection.

tritone (approved):
Looks good to me!

frankyn (approved):

tschaub:
Thanks for the work on this and for adding retries in the case of `ChunkedEncodingError`.

We're seeing quite a few `ChunkedEncodingError: ("Connection broken: error(104, 'Connection reset by peer')", error(104, 'Connection reset by peer'))` failures while paging through `bucket.list_blobs()` queries to GCS, and hoping that retries on `ChunkedEncodingError` will improve the situation.

Are there regularly scheduled releases of this package?  Or is there another way we can get a sense for when this change might be released?

andrewsg:
@tschaub Working on a release today.

andrewsg:
Correction, it's Friday so we'll target Monday.

PR #201, opened 2021-02-10
docs: address terminology

frankyn:
Address terminology.


frankyn:
@busunkim96 do you know why `Kokoro docs-presubmit` would fail? There's no result in kokoro and I've ran both lint and docs through nox locally with both passing.

frankyn:
Missing a config per @busunkim96, ignoring docs-presubmit. Just need a review.

andrewsg (approved):

PR #216, opened 2021-05-08
feat: allow RetryStrategy to be configured with a custom initial wait and multiplier

andrewsg:
This is a backwards-compatible change to allow the python-storage client to configure retries in media calls similarly to how it configures retries in API calls.

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -189,7 +189,9 @@ def wait_and_retry(func, get_status_code, retry_strategy):
 
             return response
 
-        base_wait, wait_time = calculate_retry_wait(base_wait, retry_strategy.max_sleep, retry_strategy.multiplier)
+        base_wait, wait_time = calculate_retry_wait(
+            base_wait, retry_strategy.max_sleep, retry_strategy.multiplier
+        )
 
         num_retries += 1
         total_sleep += wait_time
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -106,7 +106,7 @@ def require_status_code(response, status_codes, get_status_code, callback=do_not
     return status_code
 
 
-def calculate_retry_wait(base_wait, max_sleep):
+def calculate_retry_wait(base_wait, max_sleep, multiplier=2.0):
     """Calculate the amount of time to wait before a retry attempt.
 
     Wait time grows exponentially with the number of attempts, until
@@ -117,15 +117,16 @@ def calculate_retry_wait(base_wait, max_sleep):
 
     Args:
         base_wait (float): The "base" wait time (i.e. without any jitter)
-            that will be doubled until it reaches the maximum sleep.
+            that will be multiplied until it reaches the maximum sleep.
         max_sleep (float): Maximum value that a sleep time is allowed to be.
+        multiplier (float): Multiplier to apply to the base wait.
 
     Returns:
         Tuple[float, float]: The new base wait time as well as the wait time
         to be applied (with a random amount of jitter between 0 and 1 seconds
         added).
     """
-    new_base_wait = 2.0 * base_wait
+    new_base_wait = multiplier * base_wait
     if new_base_wait > max_sleep:
         new_base_wait = max_sleep
 
@@ -157,7 +158,8 @@ def wait_and_retry(func, get_status_code, retry_strategy):
     """
     total_sleep = 0.0
     num_retries = 0
-    base_wait = 0.5  # When doubled will give 1.0
+    # base_wait will be multiplied by the multiplier on the first retry.
+    base_wait = float(retry_strategy.initial_delay) / retry_strategy.multiplier
 
     # Set the retriable_exception_type if possible. We expect requests to be
     # present here and the transport to be using requests.exceptions errors,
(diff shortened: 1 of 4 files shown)


google/resumable_media/_helpers.py
```diff
@@ -157,7 +158,8 @@ def wait_and_retry(func, get_status_code, retry_strategy):
     """
     total_sleep = 0.0
     num_retries = 0
-    base_wait = 0.5  # When doubled will give 1.0
+    # base_wait will be multiplied by the multiplier on the first retry.
+    base_wait = float(retry_strategy.initial_delay) / retry_strategy.multiplier
```
tseaver, line 162: This change needs coverage in `tests/unit/test__helpers.py` (presumably by setting non-default values to the `initial_delay` and `multipier` attributes and checking the returned value).

google/resumable_media/common.py
```diff
@@ -142,6 +146,8 @@ def __init__(
         self.max_sleep = max_sleep
         self.max_cumulative_retry = max_cumulative_retry
         self.max_retries = max_retries
+        self.initial_delay = initial_delay
+        self.multiplier = multiplier
```
tseaver, line 150: This bit needs explicit tests in `tests/unit/test_common` for the ctor defaults and for explicitly-passed values.

tseaver (changes requested):

andrewsg:
@tseaver thanks, PTAL

tests/unit/test__helpers.py
```diff
@@ -203,6 +203,38 @@ def test_success_with_retry(self, randint_mock, sleep_mock):
         sleep_mock.assert_any_call(2.625)
         sleep_mock.assert_any_call(4.375)
 
+    @mock.patch(u"time.sleep")
+    @mock.patch(u"random.randint")
+    def test_success_with_retry_custom_delay(self, randint_mock, sleep_mock):
+        randint_mock.side_effect = [125, 625, 375]
```
tseaver, line 209: Mapping these values onto the expected `sleep` calls below took a fair bit of mental gymnastics.  Maybe we could make it more legible, e.g.:

```python
initial_delay = 3.0  # sec
multiplier = 4
random_jitter = randint_mock.side_effect = [125, 625, 375]  # ms
expected_sleeps = [
    initial_delay * multiplier * index + jitter
    for index, jitter in enumerate(random_jitter)
]
```

And then below:

```python
assert sleep_mock.assert_has_calls([
    mock.call(expected_sleep) for expected_sleep in expected_sleeps
])
```

tseaver (changes requested):

tests/unit/test__helpers.py
```diff
@@ -195,6 +203,38 @@ def test_success_with_retry(self, randint_mock, sleep_mock):
...
+
+        assert randint_mock.call_count == 3
+        assert randint_mock.mock_calls == [mock.call(0, 1000)] * 3
+
+        assert sleep_mock.call_count == 3
+        sleep_mock.assert_any_call(3.125)
+        sleep_mock.assert_any_call(12.625)
+        sleep_mock.assert_any_call(48.375)
+
+
```
tseaver, line 237: `flake8` chokes here:
> tests/unit/test__helpers.py:238:5: E303 too many blank lines (2)

tests/unit/test__helpers.py
andrewsg: In the interest of being uniform across all similar tests, I'll take a shot at clarifying this with comments first.

tests/unit/test__helpers.py
andrewsg: That's very strange as I ran blacken and lint before submitting; I'll take a look.

tests/unit/test__helpers.py
andrewsg: PTAL; do these comments clarify sufficiently?

tseaver: The comments do clarify the intent.  I'm not sure why this module prefers the count + `assert_any_call` bit is preferred over the stronger `assert_has_calls` (which tests order by default), but we can let that go here.

tseaver (approved):

tests/unit/test__helpers.py
```diff
@@ -195,6 +203,41 @@ def test_success_with_retry(self, randint_mock, sleep_mock):
         sleep_mock.assert_any_call(2.625)
         sleep_mock.assert_any_call(4.375)
 
+    @mock.patch(u"time.sleep")
+    @mock.patch(u"random.randint")
+    def test_success_with_retry_custom_delay(self, randint_mock, sleep_mock):
```
tritone, line 208: Nice test!

google/resumable_media/_helpers.py
```diff
@@ -187,7 +189,9 @@ def wait_and_retry(func, get_status_code, retry_strategy):
 
             return response
 
-        base_wait, wait_time = calculate_retry_wait(base_wait, retry_strategy.max_sleep)
+        base_wait, wait_time = calculate_retry_wait(
+            base_wait, retry_strategy.max_sleep, retry_strategy.multiplier
+        )
```
tritone, line 194: Kind of surprising that this is all calculated in the library rather than using a helper (in Go we use gax) but obviously nothing to address here.

tritone (approved):
Looks good, thanks! @frankyn FYI as well.

google/resumable_media/_helpers.py
andrewsg: Strong agree. We have such a helper for python-storage but because the transport layer is different I couldn't simply replace the implementation here with that helper. In the future I'd like to do a bit more work to bring resumable media into the fold in terms of unifying its behavior with python-storage and BQ.

PR #229, opened 2021-06-16
fix!: drop Python 2.7 support

### LINKED ISSUE #228: Drop Python 2.7 support
This would be semver major.

We should verify that clients which cannot move already pin google.resumable-media < 2.0.0dev directly, or transitively.

Extra cleanups:

- ~~Drop `mock` wrapper (use `from unittest import mock` directly).~~ We cannot drop `mock` because Python 3.6 / 3.7 do not have `unittest.mock.AsyncMock` in the stdlib.
- [x] Drop `six`.
- [x] Drop `u"` prefixes for text strings.

tseaver:
- Drop `six`.
- Drop `u"` prefixes for text strings.

Release-As: 2.0.0b1

Closes #228.


README.rst
```diff
@@ -16,15 +16,11 @@ support at `google._async_resumable_media`.
 
 Supported Python Versions
 -------------------------
-Python >= 3.5
-
-Deprecated Python Versions
---------------------------
-Python == 2.7. Python 2.7 support will be removed on January 1, 2020.
```
plamut, line 23: Ditto here, AFAIK we should retain the "unsupported Python versions" sections and explain what the last backward-compatible library version is.

tests/system/requests/test_upload.py
```diff
@@ -14,13 +14,13 @@
 
 import base64
 import hashlib
+import http.client
 import io
 import os
+import urllib_parse
```
plamut, line 20: This should be `urllib.parse`.

google/_async_resumable_media/__init__.py
```diff
@@ -52,10 +52,10 @@
...
-    u"PERMANENT_REDIRECT",
-    u"RetryStrategy",
-    u"TOO_MANY_REQUESTS",
-    u"UPLOAD_CHUNK_SIZE",
+    "DataCorruption",
+    "InvalidResponse",
+    "PERMANENT_REDIRECT",
+    "RetryStrategy",
+    "TOO_MANY_REQUESTS",
+    "UPLOAD_CHUNK_SIZE",
```
plamut, line 60: There are also quite a few occurrences of single quoted strings with this `u` prefix.

Also, the script replaced `u` bytes in binary files, i.e. test images, which is probably why a few system tests failed.

tests/unit/test__download.py
```diff
@@ -23,8 +23,7 @@
 
 
 EXAMPLE_URL = (
-    "https://www.googleapis.com/download/storage/v1/b/"
-    "{BUCKET}/o/{OBJECT}?alt=media"
+    "https://www.googleapis.com/download/storage/v1/b/" "{BUCKET}/o/{OBJECT}?alt=media"
```
plamut, line 26: (nit) We can merge this into a single string literal.

plamut (changes requested):
A few nits, but the major thing is the incorrect import. And maybe changes in test image files need to be reverted. Otherwise looks good.

tests/system/requests/test_upload.py
tseaver: e2b99bd5edf634b21b59178843ae75aaffd27408

tests/unit/test__download.py
tseaver: 3f6d81d

README.rst
tseaver: 87c3d72

google/_async_resumable_media/__init__.py
tseaver: - `u'` literals: e7cdd35
- Images restored: 9d82941
- Systest import: e2b99bd

plamut (commented):
The `six` pin sneaked back into `constraints-3.6.txt` after merging the latest `master`. When that is removed, I think we are good to go.

**Edit:** Ugh, and that `urllib_parse` import again for some reason... (ah, it's actually still there, it was not fixed in all files)

tests/system/requests/test_upload.py
plamut: There's another file where this replacement is needed - `tests/system/requests/test_upload.py`

plamut (approved):

plamut:
I have nothing more to add, thanks for this cleanup!

tseaver:
Marked as `do not merge` pending discussion over the versioning policy next week with @busunkim96, @crwilcox, et al.

PR #230, opened 2021-06-18
chore: move non-cloud package to staging bucket

dandhlee:
For now, we will move non-cloud packages to `gs://docs-staging-v2-staging` until we will be handling non-cloud packages on `gs://docs-staging-v2`. Once that support is enabled, we'll move these back.


tseaver (approved):

dandhlee:
Will be making additional Owlbot.py updates!

dandhlee:
`docfx` is a required session under Kokoro, will have to wait until https://github.com/googleapis/sphinx-docfx-yaml/pull/48 is resolved.

dandhlee:
Seems like `GoogleDocstring` does not like how things are set up in this repo. Will investigate.

dandhlee:
For `Raises` docstring, having `~` in front of its type seems to cause failures as it expects certain types with `GoogleDocstring` method. I can't find such specifications in any of the Google's Docstring rules, but a way to handle them should be available. Converting to xref also doesn't seem to help.

Easiest fix would be to temporarily convert to remove tilde and/or away from the xref, then quickly turn it back after processing is complete.

PR #241, opened 2021-07-20
fix: add retry coverage to the streaming portion of a download

### LINKED ISSUE #240: Exception retries do not cover mid-stream errors which occur during downloads.
Retry functionality in the library only covers the initial response. While this is the most common point of failure, some errors, especially Connection Reset by Peer (which usually boils down to requests.ChunkedEncodingError), escape the retry clause because they only appear during the actual streaming/writing to file period. We will need to adjust our control flow so that retry covers both the initial response (receiving/evaluating http headers) and the streaming portion of the download.

With a 16-core GCE VM and multiple processes, I observed this issue about 10 times in 1,000,000 downloads of 0.5-1 GB files.

cojenco:
This broadens the retry coverage in downloads to cover both the initial response (receiving/evaluating http headers) and the streaming/writing to file portion of the download.  

- Modifies the original `consume()` method to an internal method `_consume()` and replace the retry-able http request to a raw http request
- Adds an internal method `consume_with_retries()` within classes `Download` and `ChunkedDownload` that will attempt to retry a callable `_consume()` until success
- Updates tests and docstrings

Fixes #240

### DIFF
--- a/google/resumable_media/_download.py
+++ b/google/resumable_media/_download.py
@@ -15,6 +15,7 @@
 """Virtual bases classes for downloading media from Google APIs."""
 
 
+import functools
 import re
 
 from six.moves import http_client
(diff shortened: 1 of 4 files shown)


andrewsg (approved):
Thanks for this awesome work!

PR #242, opened 2021-07-20
tests: unpin dependency pytest-asyncio

### LINKED ISSUE #219: Explicit version range needed for dependency pytest-asyncio?
[System test failure](https://github.com/googleapis/google-resumable-media-python/issues/214) was due to a pytest-asyncio package issue, in which the loop cleanup changes were causing test errors. To unblock PR submissions, a [fix](https://github.com/googleapis/google-resumable-media-python/pull/218) is submitted to pin pytest-asyncio to previous working version. Follow up with dependency change and whether or not we should keep it pinned with version range.

cojenco:
Previous issue seen in #218 pytest-asyncio 0.15.0 resolved in 0.15.1

Fixes #219

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,7 +40,7 @@ def unit(session):
     )
 
     # Install all test dependencies, then install this package in-place.
-    session.install("mock", "pytest", "pytest-cov", "pytest-asyncio<=0.14.0", GOOGLE_AUTH)
+    session.install("mock", "pytest", "pytest-cov", "pytest-asyncio", GOOGLE_AUTH)
     session.install("-e", ".[requests,aiohttp]", "-c", constraints_path)
 
     # Run py.test against the unit tests.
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -40,7 +40,7 @@ def unit(session):
     )
 
     # Install all test dependencies, then install this package in-place.
-    session.install("mock", "pytest", "pytest-cov", "pytest-asyncio<=0.14.0", GOOGLE_AUTH)
+    session.install("mock", "pytest", "pytest-cov", "pytest-asyncio", GOOGLE_AUTH)
```
tseaver, line 43: Given that this is a test-only dependency, maybe we should be adding `, "-c", constraints_path` here, and then maybe excluding the known-bad versions in `testing/constraints*.txt`.  Likewise for the system test session below.

cojenco:
The issue remains open, and not yet resolved in 0.15.0 nor 0.15.1, per pytest-asyncio issue#211 . Closing this PR for now.

PR #250, opened 2021-08-06
tests: test against 2.x versions of google-auth

tseaver:
Also, expand testing-only pinned range for 'google-auth' to allow 2.x.y versions.


tseaver:
Consensus from today's meeting is that testing prereleases of third-party dependencies needs to happen outside the normal presubmit path.

busunkim96 (approved):

PR #258, opened 2021-08-25
fix: check if retry is allowed after retry wait calculation

### LINKED ISSUE #256: Retries are not checked correctly in the helper method
The `RetryStrategy` object configures retries in the library. We can check if another retry is allowed using [`RetryStrategy.retry_allowed()`](https://github.com/googleapis/google-resumable-media-python/blob/a7ae87272f51ad6b6b5f7cdfb5b61c5bfa3956bf/google/resumable_media/common.py#L157). 

[`_helpers.wait_and_retry()`](https://github.com/googleapis/google-resumable-media-python/blob/a7ae87272f51ad6b6b5f7cdfb5b61c5bfa3956bf/google/resumable_media/_helpers.py#L195) handles many of the retry calls in the library and utilizes `retry_allowed()`. Given the logic in `retry_allowed()`, the check should happen after the num_retries and total_sleep incrementation.

cojenco:
[`_helpers.wait_and_retry()`](https://github.com/googleapis/google-resumable-media-python/blob/a7ae87272f51ad6b6b5f7cdfb5b61c5bfa3956bf/google/resumable_media/_helpers.py#L195) handles many of the retry calls in the library and utilizes [`retry_allowed()`](https://github.com/googleapis/google-resumable-media-python/blob/a7ae87272f51ad6b6b5f7cdfb5b61c5bfa3956bf/google/resumable_media/common.py#L157). Given the logic in `retry_allowed()`, the check should happen after the num_retries and total_sleep incrementation.

For example, if `RetryStrategy.max_retries = 0`, the check should prevent any additional retry attempts. With the previous logic, 1 retry call would be made. 

If `RetryStrategy.max_cumulative_retry = 100`, the check should prevent another retry that will cause the total sleep time to exceed 100. Moving the order of the retry_allowed check resolves this.

Fixes #256

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -192,17 +192,17 @@ def wait_and_retry(func, get_status_code, retry_strategy):
         else:
             return response
 
-        if not retry_strategy.retry_allowed(total_sleep, num_retries):
-            # Retries are exhausted and no acceptable response was received.
-            # Raise the retriable_error.
-            raise error
-
         base_wait, wait_time = calculate_retry_wait(
             base_wait, retry_strategy.max_sleep, retry_strategy.multiplier
         )
-
         num_retries += 1
         total_sleep += wait_time
+
+        # Check if (another) retry is allowed. If retries are exhausted and
+        # no acceptable response was received, raise the retriable error.
+        if not retry_strategy.retry_allowed(total_sleep, num_retries):
+            raise error
+
         time.sleep(wait_time)
 
 
--- a/google/resumable_media/common.py
+++ b/google/resumable_media/common.py
@@ -158,9 +158,10 @@ class RetryStrategy(object):
         """Check if another retry is allowed.
 
         Args:
-            total_sleep (float): The amount of sleep accumulated by the caller.
-            num_retries (int): The number of retries already attempted by
-                the caller.
+            total_sleep (float): With another retry, the amount of sleep that
+                will be accumulated by the caller.
+            num_retries (int): With another retry, the number of retries that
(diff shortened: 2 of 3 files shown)


tseaver (approved):

google/resumable_media/_helpers.py
```diff
@@ -192,17 +192,17 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
         base_wait, wait_time = calculate_retry_wait(
             base_wait, retry_strategy.max_sleep, retry_strategy.multiplier
         )
-
         num_retries += 1
         total_sleep += wait_time
+
+        # Check if (another) retry is allowed. If retries are exhausted and
+        # no acceptable response was received, raise the retriable error.
+        if not retry_strategy.retry_allowed(total_sleep, num_retries):
```
andrewsg, line 203: The docstring for `retry_strategy.retry_allowed()` says that the num_retries is supposed to be the number of retries used so far. After thinking about it, I agree with your changes here and think that docstring is wrong. We should modify the docstring to reflect this change.

andrewsg (changes requested):
This looks good but I'm concerned at why there aren't more tests we have to change to reflect this. Do we not have any tests that test the case where retry strategy max_retries is used, instead of the timer-based settings? Do we need to add one, as a regression test?

google/resumable_media/_helpers.py
cojenco: @andrewsg I've updated the docstring and added another test case. PTAL, thanks!

andrewsg (approved):

PR #261, opened 2021-08-27
fix: upload download instances are not marked invalid with retriable error codes

cojenco:
`ResumableUpload` and `ChunkedDownload` can retry when it is safe to do so (encountering retriable error codes and specify preconditions if needed). The instance should not be marked as invalid if it encounters a retriable status code.

Update helper method `require_status_code()` to only make an instance invalid with non-retriable error codes.

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -192,17 +192,17 @@ def wait_and_retry(func, get_status_code, retry_strategy):
         else:
             return response
 
-        if not retry_strategy.retry_allowed(total_sleep, num_retries):
-            # Retries are exhausted and no acceptable response was received.
-            # Raise the retriable_error.
-            raise error
-
         base_wait, wait_time = calculate_retry_wait(
             base_wait, retry_strategy.max_sleep, retry_strategy.multiplier
         )
-
         num_retries += 1
         total_sleep += wait_time
+
+        # Check if (another) retry is allowed. If retries are exhausted and
+        # no acceptable response was received, raise the retriable error.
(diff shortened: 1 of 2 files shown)


google/resumable_media/_helpers.py
```diff
@@ -192,17 +193,17 @@ def wait_and_retry(func, get_status_code, retry_strategy):
...
         )
-
         num_retries += 1
         total_sleep += wait_time
+
+        # Check if (another) retry is allowed. If retries are exhausted and
+        # no acceptable response was received, raise the retriable error.
+        if not retry_strategy.retry_allowed(total_sleep, num_retries):
+            raise error
+
```
tseaver, line 206: This part of the change duplicates PR #261.  Can you rebase against `main` instead?

google/resumable_media/_helpers.py
```diff
@@ -95,7 +95,8 @@ def require_status_code(response, status_codes, get_status_code, callback=do_not
     """
     status_code = get_status_code(response)
     if status_code not in status_codes:
-        callback()
+        if status_code not in common.RETRYABLE:
+            callback()
```
tseaver, line 99: We need an additional test (or tests) below for the case being fixed here:  it should have a line like:

```python
        callback.assert_not_called()
```

tseaver (changes requested):

google/resumable_media/_helpers.py
cojenco: Added test cases for the retryable status codes. PTAL, thanks!

tseaver (approved):

andrewsg (approved):
Elegant solution!

PR #264, opened 2021-09-02
fix: pin google-crc32c to 1.1.2 to mitigate issue on OS X Big Sur

andrewsg:
This pin avoids a google-crc32c segmentation fault for OS X Big Sur users that eluded CI (which is running an older version of OS X). Pin can be removed once google-crc32c is fixed for OS X Big Sur.


cojenco (approved):
Thanks for the mitigation!

dandhlee (approved):

PR #267, opened 2021-09-08
fix: un-pin google-crc32c

tseaver:
This reverts commit 9fa344f42a99db1af27b8ca126a2ea6b3c01d837.


andrewsg (approved):

andrewsg (changes requested):
Hold off for a sec, there is an automated system that manages these that we silenced earlier. Let me try to get that back online instead of doing it manually.

andrewsg (approved):
Okay, looks like that system was only silenced for that one release. We're good.

PR #270, opened 2021-09-10
fix: add REQUEST_TIMEOUT 408 as retryable code

### LINKED ISSUE #271: Retryable codes missing 408 Request Timeout
HTTP status code 408's [Request Timeout ](https://cloud.google.com/storage/docs/json_api/v1/status-codes#408_Request_Timeout)can be handled with exponential backoff and is retryable. `http.client.REQUEST_TIMEOUT` should be included in retryable codes.

cojenco:
This adds `http.client.REQUEST_TIMEOUT` (408) to retryable codes.

Per team discussion with the GCS FnD team, [HTTP status code 408's Request Timeout](https://cloud.google.com/storage/docs/json_api/v1/status-codes#408_Request_Timeout) can be handled with exponential backoff and is retryable. An uploadBrokenConnection will cause a 408. 408 is currently included in [retryable codes in python-storage](https://github.com/googleapis/python-storage/blob/a0f05a6a7176e531af1768ffad12415dbef589b2/google/cloud/storage/retry.py#L42), and should also be included in the resumable media package.

Fixes #271

### DIFF
--- a/tests/unit/test__helpers.py
+++ b/tests/unit/test__helpers.py
@@ -198,14 +198,9 @@ class Test_wait_and_retry(object):
     @mock.patch("time.sleep")
     @mock.patch("random.randint")
     def test_success_with_retry(self, randint_mock, sleep_mock):
-        randint_mock.side_effect = [125, 625, 375]
+        randint_mock.side_effect = [125, 625, 375, 0, 500, 500]
 
-        status_codes = (
-            http.client.INTERNAL_SERVER_ERROR,
-            http.client.BAD_GATEWAY,
-            http.client.SERVICE_UNAVAILABLE,
-            http.client.NOT_FOUND,
-        )
+        status_codes = common.RETRYABLE + (http.client.NOT_FOUND,)
         responses = [_make_response(status_code) for status_code in status_codes]
 
         def raise_response():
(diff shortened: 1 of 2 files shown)


tseaver (changes requested):
The change to add `http.client.REQUEST_TIMEOUT` is fine.  However, the changes to the tests are actually unrelated to that change, and unnecessary -- please revert them.

cojenco:
For `test_success_with_retry`, I've changed to retry all of the `common.RETRYABLE` codes.  Thus we can ensure that all retryable codes are successfully retried in the mocks. Reverted the 2 other unnecessary changes. PTAL, thanks!

tests/unit/test__helpers.py
```diff
@@ -222,16 +217,19 @@ def raise_response():
         assert ret_val.status_code == status_codes[-1]
         assert status_codes[-1] not in common.RETRYABLE
 
-        assert func.call_count == 4
-        assert func.mock_calls == [mock.call()] * 4
+        assert func.call_count == 7
```
tseaver, line 220: Likewise, can we set the counts here (and below) using something like `len(status_codes)` (or adding / subtracting from it)?

tseaver (changes requested):

tritone (approved):

tests/unit/test__helpers.py
cojenco: Good idea! Revised this test and other related tests in the file.

tseaver (approved):

PR #273, opened 2021-09-16
chore: clean up aliases used for older Python versions

tseaver:


google/resumable_media/common.py
```diff
@@ -25,35 +25,39 @@
 
 UPLOAD_CHUNK_SIZE = 262144  # 256 * 1024
 """int: Chunks in a resumable upload must come in multiples of 256 KB."""
-PERMANENT_REDIRECT = 308
+
+PERMANENT_REDIRECT = http.client.PERMANENT_REDIRECT
 """int: Permanent redirect status code.
 
+.. note::
+   This is a backward-compatibility alias.
```
cojenco, line 33: Thanks for the cleanup! 

Could you link me to some resources on backward-compatibility aliases in http.client? I've been checking the RFC docs and would like to know how to identify backward compatible aliases. Thanks!

cojenco (approved):

PR #282, opened 2021-10-11
fix: Include ConnectionError and urllib3 exception as retriable

andrewsg:


google/resumable_media/_helpers.py
```diff
@@ -384,10 +384,13 @@ def _get_connection_error_classes():
...
 
     import requests.exceptions
+    import urllib3.exceptions
 
     return (
         requests.exceptions.ConnectionError,
         requests.exceptions.ChunkedEncodingError,
+        urllib3.exceptions.ProtocolError,
+        ConnectionError,  # Python 3.x only, superclass of ConnectionResetError.
     )
```
tseaver, line 394: Given that we are adding a stdlib type here, it seems like perhaps having this method raise `ImportError` is no longer "correct":  instead, perhaps the imports themselves should be conditional.

tests/unit/test__helpers.py
```diff
@@ -20,6 +20,7 @@
 import mock
 import pytest
 import requests.exceptions
+import urllib3.exceptions
```
tseaver, line 23: Having these imports outside of the `requests` subpackage is a bit manky, too.

tseaver (commented):
Rather than the changes I suggested inline, perhaps it makes more sense to move `wait_and_retry` and the associated helpers down into the `requests` sub-package?

andrewsg:
@tseaver I decided I agree with you on the architecture question; I don't remember what I was thinking when I put it in _helpers originally, but I've refactored retries to be in _request_helpers instead. I think that will resolve your feedback. PTAL

tseaver (approved):

PR #285, opened 2021-10-26
Fix: Reset stream on retries

### LINKED ISSUE #284: Stream reset on retry
Hello there, 

I am using the library, indirectly via python-storage, in maybe an unorthodox way: I am trying to implement a download timeout via a custom http adapter and response. 
So my response object is throwing pro-actively timeouts (common.InvalidResponse with http.client.REQUEST_TIMEOUT),
which is forcibly triggering the retry logic in the Download class:
https://github.com/googleapis/google-resumable-media-python/blob/d0827725085ab064169cda004a4fde65e92813f0/google/resumable_media/requests/download.py#L164

All goes well, but one thing: the final results content data from both the partial content of prematurely aborted request and the final and complete data of the second request. 
It happens because both dataset are accumulated in the `_stream` above.

Do you agree it's bug and that on retry the _stream shall be resetted (e.g. seek(0))? Do I miss a use case where it makes sense to accumulated data, or we were just assuming data would be empty anyway on exception?

If I did not miss anything I'll submit a PR to fix it.

Many thanks in advance.
Fred

fboismenu:
Fixes #284 
Reset download stream before processing a retried request.
Avoids to accumulate data from a failed request.

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -161,7 +161,14 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
             request_kwargs["stream"] = True
 
         # Wrap the request business logic in a function to be retried.
+        is_retry = False
         def retriable_request():
+
+            nonlocal is_retry
+            if is_retry and self._stream is not None:
+                self._stream.seek(0)
+            is_retry = True
+
             result = transport.request(method, url, **request_kwargs)
 
             self._process_response(result)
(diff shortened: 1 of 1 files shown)


tseaver (commented):
@fboismenu Thanks very much for the patch!  In additional to my inline comments, we also need to address the following concerns:
- In order to demonstrate the fix, we need new unit tests which provoke retries after a write to the stream, and which assert that the stream does *not* contain that initial write after finishing the download.
- We need to consider how the fix generalizes to the other two classes (`ChunkedDownload` and `RawChunkedDownload`).

cojenco:
Hi, I'm working on submitting a fix, and will link the new PR back to this. Thanks!

parthea:
Superseded by #294

PR #287, opened 2021-11-16
feat: add 'py.typed' declaration file

tseaver:
Check typing under CI using new 'mypy' nox session.

### DIFF
--- a/google/__init__.py
+++ b/google/__init__.py
@@ -19,4 +19,4 @@ try:
 except ImportError:
     import pkgutil
 
-    __path__ = pkgutil.extend_path(__path__, __name__)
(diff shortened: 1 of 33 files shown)


tseaver:
Note to reviewers:  aside from complaints about missing stubs, etc. the `mypy` check found one "real" issue:  both the sync and async versions of `ResumableUpload` redefined the signature of its inherited `_process_upload` helper method.  I fixed that issue by renaming it (in both places) to `_process_resumable_upload`.

cojenco (approved):
Thanks for adding this! I'll add "py.typed" with changes to python-storage once python 2 deprecation is finalized.

PR #290, opened 2021-12-08
feat: add support for signed resumable upload URLs

### LINKED ISSUE #289: Using a signed Resumable Upload URL fails with 403 unless content_type is empty
#### Environment details

  - Programming language: Python
  - OS: Ubuntu 20.04 in WSL2
  - Language runtime version: 3.8
  - Package version: cee416449701b72e7fd532585a1f739b02b6ab32

#### Steps to reproduce

  1. Generate a signed resumable upload url:
```
gsutil signurl -m RESUMABLE -c application/octet-stream -d 24h <key_file> gs://<bucket>/<obj>
```
  2. Run this python program, passing in the upload_url that was generated:
``` python
import io

from google.resumable_media import InvalidResponse
from google.resumable_media.requests import ResumableUpload
from requests import Session

chunk_size = 16 * 1024 * 1024  # 16MB

upload = ResumableUpload(upload_url, chunk_size)
stream = io.BytesIO(data)
metadata = {}
transport = Session()
transport.headers.update({'x-goog-resumable': 'start'})
try:
	response = upload.initiate(
		transport, stream, metadata, 'application/octet-stream')
except InvalidResponse as err:
	print(err.response.content)
	raise
print(response)
while not upload.finished:
	upload.transmit_next_chunk(transport)
stream.close()
```

You will get a 403 error from google that says `The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.`.  This is because the content_type passed by the library to the signed url is not correct.  If you instead generate the url without a content_type - it will succeed - but the object will have the content type `application/json`.  I'll send in a PR shortly with a fix that works for me.

williambrode:
Fixes #289.

### DIFF
--- a/google/resumable_media/_upload.py
+++ b/google/resumable_media/_upload.py
@@ -468,9 +468,7 @@ class ResumableUpload(UploadBase):
         parse_result = urllib.parse.urlparse(self.upload_url)
         parsed_query = urllib.parse.parse_qs(parse_result.query)
         if "x-goog-signature" in parsed_query or "X-Goog-Signature" in parsed_query:
-            headers = {
-                _CONTENT_TYPE_HEADER: content_type
-            }
+            headers = {_CONTENT_TYPE_HEADER: content_type}
         else:
             headers = {
                 _CONTENT_TYPE_HEADER: "application/json; charset=UTF-8",
--- a/google/resumable_media/_upload.py
+++ b/google/resumable_media/_upload.py
@@ -27,6 +27,7 @@ import os
 import random
 import re
 import sys
+import urllib.parse
 
 from google import resumable_media
 from google.resumable_media import _helpers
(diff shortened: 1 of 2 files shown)


williambrode:
I'm sure my code is not production-worthy yet but I wanted to create the PR to help show what the problem is from my debugging.

williambrode:
I signed the CLA.

williambrode:
This doesn't handle v2 signed urls because they use a different query param ("Signature" I believe).  But I'm not sure that's a good reason to not support V4 signed urls.  We can add this but I don't have a test for it.

tseaver:
@williambrode Thanks for the patch!

Can you double-check that the e-mail you used to create the commit (`Author: <wil*****de​@activision.com>`) matches the email(s) associated with your signed CLA?

williambrode:
@tseaver You are right my e-mail was wrong in the commit.  I've fixed it.

tseaver (changes requested):
This change looks reasonable, given the bug report, but we need to add unit tests to cover the new cases (`"x-goog-signature" in parsed_query` / ``"X-Goog-sSgnature" in parsed_query`).

Breathtender:
@williambrode just wanted to follow up with you; are you able to write unit tests for this or would you like some support from us on that aspect. I know our testing framework might be a bit of a learning curve if this is your first external contribution :)

williambrode:
Ah, I didn't quite realize that was directed at me.  I can take a look at adding the test.

Breathtender:
Haha, reading over Tres' message again, it might not have been, I think I was hasty in my pass-through.
We can definitely handle getting a unit test written unless you'd like to give that a shot.

Breathtender:
Alright, gave a go at the unit test. How's that look @tseaver ?

cojenco:
Updated the unit tests to cover signed url uploads. @tseaver PTAL, thanks!

tseaver (approved):

PR #294, opened 2022-01-18
feat: safely resume interrupted downloads

### LINKED ISSUE #284: Stream reset on retry

cojenco:
If a retryable error occurs mid-download, the download starts sending data to the stream from the `offset_of_last_byte_received` rather than starting from the beginning of the file, and resolves data integrity issues. 

- for interruped downloads, safely resume by reading from offset_of_last_byte_received using a ranged get request and include object generation URL query parameter to make sure the same object content is requested
- adds support for download instances to track information such as `object_generation` and `bytes downloaded`
- adds tests

Fixes #284

### DIFF
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -38,7 +38,6 @@ _SLOW_CRC32C_WARNING = (
     "implementation. Python 3 has a faster implementation, `google-crc32c`, "
     "which will be used if it is installed."
 )
-_CONTENT_ENCODING_HEADER = "Content-Encoding"
 _GENERATION_HEADER = "x-goog-generation"
 _HASH_HEADER = "x-goog-hash"
 _MISSING_CHECKSUM = """\
@@ -368,21 +367,6 @@ def add_query_parameters(media_url, query_params):
     new_params = {**params, **query_params}
     query = urlencode(new_params, doseq=True)
     return urlunsplit((scheme, netloc, path, query, frag))
-
-
-def _is_decompressive_transcoding(response, get_headers):
-    """Returns True if the object was served decompressed and the "Content-Encoding" header is "gzip".
-       See more at: https://cloud.google.com/storage/docs/transcoding#transcoding_and_gzip
-
-    Args:
-        response (~requests.Response): The HTTP response object.
-        get_headers (callable: response->dict): returns response headers.
-
-    Returns:
-        bool: Returns True if the "Content-Encoding" header is "gzip"; otherwise, False.
-    """
-    headers = get_headers(response)
-    return headers.get(_CONTENT_ENCODING_HEADER) == "gzip"
 
 
 class _DoNothingHash(object):
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -33,6 +33,7 @@ _SLOW_CRC32C_WARNING = (
     "implementation. Python 3 has a faster implementation, `google-crc32c`, "
     "which will be used if it is installed."
 )
+_CONTENT_ENCODING_HEADER = "Content-Encoding"
 _HASH_HEADER = "x-goog-hash"
 _GENERATION_HEADER = "x-goog-generation"
 _MISSING_CHECKSUM = """\
@@ -322,6 +323,21 @@ def _parse_generation_header(response, get_headers):
         return int(object_generation)
 
     return object_generation
+
+
+def _is_decompressive_transcoding(response, get_headers):
+    """Returns True if the object was served decompressed and the "Content-Encoding" header is "gzip".
+       See more at: https://cloud.google.com/storage/docs/transcoding#transcoding_and_gzip
+
+    Args:
+        response (~requests.Response): The HTTP response object.
+        get_headers (callable: response->dict): returns response headers.
+
+    Returns:
+        bool: Returns True if the "Content-Encoding" header is "gzip"; otherwise, False.
+    """
+    headers = get_headers(response)
+    return headers.get(_CONTENT_ENCODING_HEADER) == "gzip"
 
 
 class _DoNothingHash(object):
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -34,6 +34,7 @@ _SLOW_CRC32C_WARNING = (
     "which will be used if it is installed."
 )
 _HASH_HEADER = "x-goog-hash"
+_GENERATION_HEADER = "x-goog-generation"
 _MISSING_CHECKSUM = """\
 No {checksum_type} checksum was returned from the service while downloading {}
 (which happens for composite objects), so client-side content integrity
@@ -300,6 +301,27 @@ def _get_checksum_object(checksum_type):
         return None
     else:
         raise ValueError("checksum must be ``'md5'``, ``'crc32c'`` or ``None``")
+
+
+def _parse_generation_header(response, get_headers):
+    """Parses the generation header from an ``X-Goog-Generation`` value.
+
+    Args:
+        response (~requests.Response): The HTTP response object.
+        get_headers (callable: response->dict): returns response headers.
+
+
+    Returns:
+        Optional[long]: The object generation from the response, if it
+        can be detected from the ``X-Goog-Generation`` header; otherwise, None.
+    """
+    headers = get_headers(response)
+    object_generation = headers.get(_GENERATION_HEADER, None)
+
+    if object_generation:
+        return int(object_generation)
+
+    return object_generation
 
 
 class _DoNothingHash(object):
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -40,8 +40,8 @@ _SLOW_CRC32C_WARNING = (
     "which will be used if it is installed."
 )
 _CONTENT_ENCODING_HEADER = "Content-Encoding"
-_HASH_HEADER = "x-goog-hash"
 _GENERATION_HEADER = "x-goog-generation"
+_HASH_HEADER = "x-goog-hash"
 _MISSING_CHECKSUM = """\
 No {checksum_type} checksum was returned from the service while downloading {}
 (which happens for composite objects), so client-side content integrity
@@ -317,7 +317,6 @@ def _parse_generation_header(response, get_headers):
         response (~requests.Response): The HTTP response object.
         get_headers (callable: response->dict): returns response headers.
 
-
     Returns:
         Optional[long]: The object generation from the response, if it
         can be detected from the ``X-Goog-Generation`` header; otherwise, None.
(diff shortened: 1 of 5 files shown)


google/resumable_media/requests/download.py
```diff
@@ -86,12 +86,22 @@ def _write_to_stream(self, response):
...
-        expected_checksum, checksum_object = _helpers._get_expected_checksum(
-            response, self._get_headers, self.media_url, checksum_type=self.checksum
-        )
+        # Retrieve the expected checksum only once for the download request,
+        # then compute and validate the checksum when the full download completes.
+        # Retried requests are range requests, and there's no way to detect
+        # data corruption for that byte range alone.
+        if self._expected_checksum is None and self._checksum_object is None:
+            # `_get_expected_checksum()` may return None even if a checksum was
+            # requested, in which case it will emit an info log _MISSING_CHECKSUM.
```
tritone, line 95: What causes this case to happen? Transcoding?

google/resumable_media/requests/download.py
```diff
@@ -160,13 +171,48 @@ def consume(
         if self._stream is not None:
             request_kwargs["stream"] = True
 
+        # Assign object generation if generation is specified in the media url.
```
tritone, line 174: Would this happen via a user specifying a generation on the object? Were we not respecting this previously?

google/resumable_media/requests/download.py
```diff
@@ -160,13 +171,48 @@ def consume(
...
+                )
+
             self._process_response(result)
 
+            # With decompressive transcoding, GCS serves back the whole file regardless of the range request,
+            # thus we reset the stream position to the start of the stream.
+            # See: https://cloud.google.com/storage/docs/transcoding#range
             if self._stream is not None:
+                if _helpers._is_decompressive_transcoding(result, self._get_headers):
+                    self._stream.seek(0)
```
tritone, line 214: I didn't realize that the stream used for this method is seekable. Interesting!

google/resumable_media/_helpers.py
```diff
@@ -302,6 +310,81 @@ def _get_checksum_object(checksum_type):
...
+    else:
+        return int(object_generation[0])
+
+
+def add_query_parameters(media_url, name_value_pairs):
+    """Add query parameters to a base url.
+
+    Args:
+        media_url (str): The URL containing the media to be downloaded.
+        name_value_pairs (list[tuple[str, str]]): Names and values of the query parameters to add.
```
tritone, line 358: I think query params are more typically represented as a map or dict. Thoughts on taking that approach?

google/resumable_media/requests/download.py
```diff
@@ -285,23 +341,57 @@ def consume(
...
+            # If a generation hasn't been specified, and this is the first response we get, let's record the
+            # generation. In future requests we'll specify the generation query param to avoid data races.
+            if self._object_generation is None:
+                self._object_generation = _helpers._parse_generation_header(
+                    result, self._get_headers
+                )
 
             self._process_response(result)
 
+            # With decompressive transcoding, GCS serves back the whole file regardless of the range request,
```
tritone, line 388: Wondering if this should be highlighted as a shortcoming in the decompressive transcoding docs-- not being able to resume a download may be costly.

tritone (commented):
Couple thoughts, generally looking really good to me!

General comment, can you give some more details about how you tested this out with the emulator?

cojenco:
> Couple thoughts, generally looking really good to me!
> 
> General comment, can you give some more details about how you tested this out with the emulator?

Thanks for the review! I've added data integrity checks and test cases to the retry conf test ([open PR](https://github.com/googleapis/python-storage/pull/702)). The changes in this PR are tested against the testbench using above-mentioned tests.

Before the changes, conformance tests fail as below. The conf tests pass running locally against the changes made in this PR.
```
  File "/tmpfs/src/github/python-storage/tests/conformance/test_conformance.py", line 93, in blob_download_as_bytes
    assert stored_contents == payload
AssertionError: assert b'ThisThisThi... text file.\n' == b'This is a s... text file.\n'
  At index 4 diff: b'T' != b' '
  Full diff:
  - b'This is a simple text file.\n'
  ?       ^
  + b'ThisThisThis is a simple text file.\n'
  ?       ^^  +++++++
```

```
=========================== short test summary info ============================
FAILED tests/conformance/test_conformance.py::test-S8-storage.objects.get-blob_download_to_filename-0
FAILED tests/conformance/test_conformance.py::test-S8-storage.objects.get-client_download_blob_to_file-0
FAILED tests/conformance/test_conformance.py::test-S8-storage.objects.get-blob_download_as_bytes-0
FAILED tests/conformance/test_conformance.py::test-S8-storage.objects.get-blobreader_read-0
FAILED tests/conformance/test_conformance.py::test-S8-storage.objects.get-blob_download_as_text-0
5 failed, 555 passed, 5 skipped, 7 warnings in 287.55s (0:04:47)
nox > Command py.test -n auto --quiet tests/conformance failed with exit code 1
nox > Session conftest_retry-3.8 failed.
```

google/resumable_media/requests/download.py
cojenco: This is due to retried requests being range requests. For range requests, as noted [here](https://cloud.google.com/storage/docs/hashes-etags#integrity), there's no way to detect data corruption for that byte range alone. 

Therefore, here we retrieve the expected checksum/checksum object only once for the initial download request. Then we calculate and validate the checksum when the download completes.

google/resumable_media/requests/download.py
cojenco: Yep this would happen via a user specifying a generation on the object. Previously, we've been respecting that only through `download.media_url`

A property `download._object_generation` is added. It records the object generation from either (1) generation query param from the media_url, or (2) the object generation from the initial response header. This specific line of code does (1) and retrieves it from the media_url

P.S. It's tricky in how limited information is passed from python-storage to resumable-media-python. A resumable-media-python `download` instance only knows the specified object generation from its `media_url`, and the "object" itself isn't pertained in a download.

google/resumable_media/_helpers.py
cojenco: That makes sense, updating now

google/resumable_media/requests/download.py
cojenco: It's mentioned in the very bottom [section](https://cloud.google.com/storage/docs/transcoding#range) of the decompressive transcoding docs. I agree we can add notes on how retries may be impacted in this sense.

google/resumable_media/requests/download.py
cojenco: I ran a few experiments and the stream is seekable, though I'm not 100% sure if we can guarantee it to be seekable. 

The `requests.Response` object seems to be immutable, and [`iter_content()`](https://github.com/googleapis/google-resumable-media-python/blob/b529b780253bed9e5619e4c5eb2d612e975d874c/google/resumable_media/requests/download.py#L102) iterates from the beginning of the response. I wasn't able to find anything similar in Python to the io.CopyN approach in Go.

@frankyn @andrewsg @tseaver  Would love to hear your ideas on how to avoid writing duplicate data to the stream for decompressive transcoding interruptions. An example scenario could be:

--> initial download request
--> some bytes have already been written to stream
!!! connection error mid-stream
--> resumes download and sends a retried range request
--> the response serves the entire requested object, instead of the partial contents requested 
(decompressive transcoding does not support range requests)

andrewsg: Sorry, to clarify, the remote server will _not_ reject a request with an unsupported range parameter, but instead will simply serve the wrong (complete) range?

What the heck? That sounds catastrophically broken

andrewsg: I only see two solutions: just error out on seek() failing (with a try/except to transform the normal error message into something useful), or simply throw away all bytes up to tell() without writing them to the stream at all, and then write remaining bytes to the stream.

I prefer the simpler solution of the former, absent any customer demand for the more complex solution. A user deliberately passing in an un-seekable stream is a rare case and we don't need to write a bunch of hard to test code to support it unless it's obvious people are really doing that.

cojenco: > Sorry, to clarify, the remote server will _not_ reject a request with an unsupported range parameter, but instead will simply serve the wrong (complete) range?
> 
> What the heck? That sounds catastrophically broken

That's right... the server does _not_ reject the request. To quote the [doc](https://cloud.google.com/storage/docs/transcoding#range)

_When transcoding occurs, if the request for the object includes a Range header, that header is silently ignored._

cojenco: > I only see two solutions: just error out on seek() failing (with a try/except to transform the normal error message into something useful), or simply throw away all bytes up to tell() without writing them to the stream at all, and then write remaining bytes to the stream.
> 
> I prefer the simpler solution of the former, absent any customer demand for the more complex solution. A user deliberately passing in an un-seekable stream is a rare case and we don't need to write a bunch of hard to test code to support it unless it's obvious people are really doing that.

That makes sense, thanks! The complex solution would be difficult to test. I'll add a clear message if seek() errors out.

google/resumable_media/_helpers.py
```diff
@@ -302,6 +310,81 @@ def _get_checksum_object(checksum_type):
...
+
+    Args:
+        response (~requests.Response): The HTTP response object.
+        get_headers (callable: response->dict): returns response headers.
+
+    Returns:
+        bool: Returns True if the "Content-Encoding" header is "gzip"; otherwise, False.
+    """
+    headers = get_headers(response)
+    return headers.get(_CONTENT_ENCODING_HEADER) == "gzip"
```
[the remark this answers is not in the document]
cojenco: re: retrying a decompressive transcoding download
Based on our internal team discussion, I'm thinking at this point of the time to
(1) check for "X-guploader-response-body-transformations" from the response headers 
(2) add a note in the docs about the restrictions of retrying decompressive transcoding downloads
(3) open an issue that links to internal FR b/218348047; add full retry support if the FR is fulfilled

@frankyn @tritone how does that sound?

tritone:
This is looking really good in general. Based on offline discussion I would recommend moving the decompressive transcoding feature to a TODO and moving ahead with the rest of this PR. There may be some details that take a while to resolve for transcoding and it's important that we still move ahead with the rest of this PR which is a major fix to retry logic for downloads.

andrewsg (approved):
LGTM pending @tritone comment resolutions. Thank you!

cojenco:
Thanks Chris and Andrew! I've moved the transcoding feature, tracking in #303

tritone (approved):

PR #295, opened 2022-01-25
fix: add user-agent on requests

unforced:
Work in progress. Still need to change a number of tests, especially on uploads.

### DIFF
--- a/google/resumable_media/_download.py
+++ b/google/resumable_media/_download.py
@@ -16,6 +16,7 @@
 
 
 import http.client
+import pkg_resources
 import re
 
 from google.resumable_media import _helpers
@@ -58,6 +59,8 @@ class DownloadBase(object):
         self.end = end
         if headers is None:
             headers = {}
+        version = pkg_resources.get_distribution('google-resumable-media').version
+        headers['User-Agent'] = "google-resumable-media-python/{}".format(version)
         self._headers = headers
         self._finished = False
         self._retry_strategy = common.RetryStrategy()
--- a/google/resumable_media/_download.py
+++ b/google/resumable_media/_download.py
@@ -59,9 +59,7 @@ class DownloadBase(object):
         self.end = end
         if headers is None:
             headers = {}
-        version = pkg_resources.get_distribution('google-resumable-media').version
-        headers['User-Agent'] = "google-resumable-media-python/{}".format(version)
-        self._headers = headers
+        self._headers = _helpers._base_headers(headers)
         self._finished = False
         self._retry_strategy = common.RetryStrategy()
 
--- a/google/resumable_media/_download.py
+++ b/google/resumable_media/_download.py
@@ -16,7 +16,6 @@
 
 
 import http.client
-import pkg_resources
 import re
 
 from google.resumable_media import _helpers
--- a/google/resumable_media/_helpers.py
+++ b/google/resumable_media/_helpers.py
@@ -20,6 +20,7 @@ import base64
 import hashlib
 import logging
 import random
+import pkg_resources
 import warnings
 
 from google.resumable_media import common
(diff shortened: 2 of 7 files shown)


google/resumable_media/_download.py
```diff
@@ -58,6 +59,8 @@ def __init__(self, media_url, stream=None, start=None, end=None, headers=None):
         self.end = end
         if headers is None:
             headers = {}
+        version = pkg_resources.get_distribution('google-resumable-media').version
+        headers['User-Agent'] = "google-resumable-media-python/{}".format(version)
```
frankyn, line 63: can this be in a helper method to improve maintainability?

frankyn, line 63: Right now we internally track `google-python/{}`; we might need to update this to `google-python-resumable-media/{}`, my only queestion is if we need to create another UserAgent instead of just using `google-python/{}`

frankyn (changes requested):

google/resumable_media/_download.py
Breathtender: Yep, we can do that.

Breathtender: That sounds reasonable. @andrewsg any thoughts?

Breathtender: I think google-python might actually make the most sense, since this is mostly called from inside of python-storage and so giving it it's own user agent could be confusing.

frankyn: I did see that https://github.com/googleapis/google-resumable-media-python/pull/84#issuecomment-525827813, the transport passed into this package is expected to have the user-agent. I'm thinking we may want to make this change within python-storage only; however we would lose tracking for this package if used outside of python-storage. 

I raised the question in Yoshi Chat room because I'm a bit confused.

Breathtender: Ha, andrew and I are sharing with you in much of that confusion. Perhaps we can all hop on a call and discuss.

andrewsg: python-storage uses gcloud-python, not google-python. Both this library and python-storage include version numbers, and if we use the same header those version numbers will conflict. I'll continue discussion in chat.

google/resumable_media/_helpers.py
```diff
@@ -45,6 +47,13 @@ def do_nothing():
     """Simple default callback."""
 
 
+def _base_headers(headers):
+    version = pkg_resources.get_distribution("google-resumable-media").version
+    headers["User-Agent"] = "gcloud-python/{}-resumable-media".format(version)
+    headers["X-Goog-Api-Client"] = "gl-python/" + platform.python_version() + " gccl/{}"
```
andrewsg, line 53: In python-storage, X-Goog-Api-Client comes out identical to User-Agent which I assumed was the intended design. But that isn't happening in the place in the code I thought it happened, which was storage/_http.py, so I am confused how that came to be.

I think (?) it should be identical to user agent in this case as well.

andrewsg (changes requested):

frankyn (approved):
Verified that this is propagating as expected.

andrewsg (approved):
Thanks!

tswast:
This is causing some test regressions in BigQuery. See https://source.cloud.google.com/results/invocations/ca676754-ee34-49ff-b6dd-b680e3dab10c/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log and https://source.cloud.google.com/results/invocations/ed53c1b0-a688-4aca-81d5-4bdc53e45e8e/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fprerelease-deps-3.8/log

google/resumable_media/_helpers.py
```diff
@@ -45,6 +46,13 @@ def do_nothing():
     """Simple default callback."""
 
 
+def _base_headers(headers):
+    version = pkg_resources.get_distribution("google-resumable-media").version
+    headers["User-Agent"] = "gcloud-python/{}-resumable-media".format(version)
```
tswast, line 51: This should not blindly override headers. At most it should append this to the end of an existing header.

I recommend using `setdefault()` or similar logic next time.

PR #309, opened 2022-03-01
fix: include existing headers in prepare request

unforced:
This is a needed change to allow us to properly set user agent from the python-storage library, for multi-chunked uploads.

### DIFF
--- a/tests/unit/test__upload.py
+++ b/tests/unit/test__upload.py
@@ -617,11 +617,11 @@ class TestResumableUpload(object):
         assert headers == expected_headers
 
     def test__prepare_request_success_with_headers(self):
-        headers = {"cannot": "touch this"}
+        headers = {"keep": "this"}
         new_headers = self._prepare_request_helper(headers)
         assert new_headers is not headers
         expected_headers = {
-            "cannot": "touch this",
+            "keep": "this",
             "content-range": "bytes 0-32/33",
             "content-type": BASIC_CONTENT,
         }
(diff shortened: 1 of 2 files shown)


tswast (approved):
As talked about in a Chat thread, it appears this reverts https://github.com/googleapis/google-resumable-media-python/commit/fd4733819f9775973f91048603ce08e959fbd375

From what I can tell, that was done purely as an optimization, so I don't see this causing any problems.

I do recommend running this patch against the GCS system tests just to be sure.

tests/unit/test__upload.py
```diff
@@ -621,12 +621,11 @@ def test__prepare_request_success_with_headers(self):
         new_headers = self._prepare_request_helper(headers)
         assert new_headers is not headers
         expected_headers = {
+            "cannot": "touch this",
```
frankyn, line 624: This test is very subtle and relies on https://github.com/googleapis/google-resumable-media-python/blob/cee416449701b72e7fd532585a1f739b02b6ab32/tests/unit/test__upload.py#L581 to pass in new headers when calling `self._upload_in_flight(data, headers=headers, checksum=checksum)`. IIUC, the test is to prevent any regressions from future changes (super not clear by the name of the test).

What I don't understand is what changed when `**self._headers` is added so it can be tested. @Breathtender could add a test that validates the behavior change intended by adding `**self._headers`?

frankyn (changes requested):

frankyn (approved):

tests/unit/test__upload.py
Breathtender: Oh that's a good change lol

andrewsg (approved):

PR #314, opened 2022-03-04
fix: append existing headers in prepare_initiate_request

unforced:
Big uggghhhhh, but the last change missed a necessary place for headers to be appended and I missed that in the testing. This time, double checked and all things pass, and so this is requiring cutting another quick release but it will allow for that change to come fully through.

### DIFF
--- a/google/resumable_media/_upload.py
+++ b/google/resumable_media/_upload.py
@@ -468,8 +468,10 @@ class ResumableUpload(UploadBase):
         parse_result = urllib.parse.urlparse(self.upload_url)
         parsed_query = urllib.parse.parse_qs(parse_result.query)
         if "x-goog-signature" in parsed_query or "X-Goog-Signature" in parsed_query:
+            # Deconstruct **self._headers first so that content type defined here takes priority
             headers = {**self._headers, _CONTENT_TYPE_HEADER: content_type}
         else:
+            # Deconstruct **self._headers first so that content type defined here takes priority
             headers = {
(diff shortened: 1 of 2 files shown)


andrewsg (approved):

frankyn (approved):

google/resumable_media/_upload.py
```diff
@@ -468,9 +468,10 @@ def _prepare_initiate_request(
         parse_result = urllib.parse.urlparse(self.upload_url)
         parsed_query = urllib.parse.parse_qs(parse_result.query)
         if "x-goog-signature" in parsed_query or "X-Goog-Signature" in parsed_query:
-            headers = {_CONTENT_TYPE_HEADER: content_type}
+            headers = {**self._headers, _CONTENT_TYPE_HEADER: content_type}
```
frankyn, line 471: Are there tests that you can update as well?

frankyn (changes requested):

google/resumable_media/_upload.py
Breathtender: Done. There was already a test that ensured they were merged together, what we changed was putting these ones first, to ensure the content_type and x-upload-content-type described here took priority. And so I just adjusted the prepare_initiate_request_with_headers by adding a content-type and ensuring it was overwritten.

frankyn (approved):
thanks @Breathtender for getting this in!

PR #329, opened 2022-05-19
fix: Use unittest.mock

### LINKED ISSUE #328: Use unittest.mock if available rather than deprecated mock module
#### Environment details

  - OS type and version: Fedora 36
  - Python version: 3.10.4
  - pip version: 22.0.4
  - `google-resumable-media` version: main branch

#### Steps to reproduce

  1. Run tests on Python 3.10 without `mock` installed
  2. Test fail due to `mock` import

I'll open a PR to fix in a moment.

major:
The `mock` module is deprecated in recent Python versions.

Signed-off-by: Major Hayden <major@mhtx.net>

Fixes #328 🦕

### DIFF
--- a/tests/system/requests/test_upload.py
+++ b/tests/system/requests/test_upload.py
@@ -20,7 +20,7 @@ import os
 import urllib.parse
 
 import pytest  # type: ignore
-import mock
+from unittest import mock
 
 from google.resumable_media import common
 from google import resumable_media
--- a/tests/unit/requests/test__helpers.py
+++ b/tests/unit/requests/test__helpers.py
@@ -14,7 +14,7 @@
 
 import http.client
 
-import mock
+from unittest import mock
 import pytest  # type: ignore
 
 import requests.exceptions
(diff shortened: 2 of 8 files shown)


tests/system/requests/test_upload.py
```diff
@@ -20,7 +20,10 @@
 import urllib.parse
 
 import pytest  # type: ignore
-import mock
+try:
```
andrewsg, line 23: Thank you for doing this! I'm pleased to say we don't need the try/except anymore, as we have a separate Python 2.7 branch and so all supported versions of the client library should support unittest.mock. Just the bare `from unittest import mock` should be enough.

andrewsg (changes requested):

tests/system/requests/test_upload.py
major: Thanks! Will update it now.

andrewsg (approved):
Thanks!

andrewsg:
Unfortunately, tests are failing now due to the change to async. Go ahead and remove any changes to the async directory and I will clean it up later along with the rest of our experimental async infrastructure.

major:
@andrewsg Done!

PR #346, opened 2022-08-15
feat: handle interrupted downloads with decompressive transcoding

### LINKED ISSUE #303: Handle retries in decompressive transcoding downloads
Add special handling for resuming an interrupted download in which transcoding occured

Pending on internal issue b/218894651 and b/218348047

cojenco:
Handle interrupted downloads with decompressive transcoding
- distinguish if transcoding occured by the newly introduced headers `x-goog-stored-content-encoding` and `x-goog-stored-content-length` in the JSON API 🎉
- check edge case in which the object is served as-is without decompression along with response header `content-encoding: gzip`
- reset the stream position for retried range requests if decompressive transcoding occurred (GCS serves back the entire file)

Fixes #303 🦕


google/resumable_media/_helpers.py
```diff
@@ -369,6 +371,19 @@ def add_query_parameters(media_url, query_params):
...
+    """Returns True if the object was served decompressed and the "x-goog-stored-content-encoding"
+    header is "gzip". See more at: https://cloud.google.com/storage/docs/transcoding#transcoding_and_gzip
+    Args:
+        response (~requests.Response): The HTTP response object.
+        get_headers (callable: response->dict): returns response headers.
+    Returns:
+        bool: Returns True if the "x-goog-stored-content-encoding" header is "gzip"; otherwise, False.
+    """
+    headers = get_headers(response)
+    return headers.get(_STORED_CONTENT_ENCODING_HEADER) == "gzip"
```
frankyn, line 384: You also want to check if `content-encoding` was not in the response headers with the value `gzip`. There's an edge case in that GCS return compressed data with `content-encoding: gzip`.

frankyn (changes requested):

google/resumable_media/_helpers.py
cojenco: That's right. Thanks Frank for catching that. I've added the check and a test alongside.

frankyn (approved):

PR #386, opened 2023-07-18
fix: Add google-auth to aiohttp extra

### LINKED ISSUE #384: Unit tests fail with `E   ModuleNotFoundError: No module named 'google.auth'`
Unit tests in https://github.com/googleapis/google-resumable-media-python/pull/383 are failing with `E   ModuleNotFoundError: No module named 'google.auth'`

See build log [here](https://github.com/googleapis/google-resumable-media-python/actions/runs/5577531514/jobs/10190414428?pr=383). 

Recently google auth added support for native namespace packages. This failure only appears with `google-auth>2.21.0`

clundin25:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-resumable-media-python/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #384


parthea (commented):
LGTM. Alternatively, we could add `google-auth` as an extra, and add conditional logic here to raise an exception to let users know that they need to install the `google-auth` extra.
https://github.com/googleapis/google-resumable-media-python/blob/main/google/_async_resumable_media/requests/_request_helpers.py#L26

clundin25:
@parthea an extra sounds best to me.

cojenco:
+1 to adding an extra instead as the async support is experimental and subject to change. 

Additionally, we're also planning to integrate this library into python-storage in the near future. I'm trying to understand how the namespace packaging changes would impact this if any; cc @andrewsg

andrewsg:
I would prefer extra, or else removing async testing. Async support is currently theoretical and not documented or supported.

clundin25:
Out of curiosity, is there anyone working on stabilizing the Async API?

andrewsg:
We don't have it on our roadmap at present as it would require a complete rewrite of python-storage, as well as an additional rewrite of this library in order to avoid having two separate versions of the library that need to be maintained simultaneously. This doesn't mean that async won't happen at all, just that it's not scheduled as of yet.

clundin25:
Okay awesome. Are you a good point of contact for this? I have some questions I want to take offline.

andrewsg:
Yep, feel free to reach out.

clundin25:
I will fix the error tomorrow. Looks like some of the sync tests also use google auth

parthea (approved):

PR #435, opened 2024-05-14
fix: add a check for partial response data

cojenco:
Handle corner cases of incomplete downloads with a 200 response code. These are rare cases when a download is prematurely terminated due to network stability or bandwidth throttling etc, but does not result in connection reset errors. The library will 
- (1) compare the # bytes read vs x-goog-stored-content-length header
- (2) raise a retry-able ConnectionError in the aim to trigger a retry

Add more context to checksum mismatch error message that will help with debugging
- bytes read from the download request (tracked within the Download object)
- `X-Goog-Stored-Content-Length` value from the response header
- `X-Goog-Stored-Content-Encoding` value from the response header

Fixes internal b/278011003

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -133,6 +133,23 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
                 self._bytes_downloaded += len(chunk)
                 local_checksum_object.update(chunk)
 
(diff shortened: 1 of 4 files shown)


andrewsg (approved):
Thanks!

danielduhh:
Is this good to merge? @cojenco

PR #447, opened 2024-07-31
Fix range read when interrupted for self.start is not None

### LINKED ISSUE #446: Incorrect start offset when range reads are interrupted
#### Environment details

 GCP A100 machines
(unclear about exact software, as this bug was countered internally on Google infra when running on XManager Cloud)

#### Steps to reproduce

Given a range read of `start` to `end` bytes, if the read is interrupted, the code [here](https://github.com/googleapis/google-resumable-media-python/blob/5d2091d78fb71e0d63254c5b731afb05af432104/google/resumable_media/requests/download.py#L429) will make another GET request at the wrong start offset, as it should be `self.start + self._bytes_downloaded` if `self.start is not None`. Currently, the downloader will return `end + 1` bytes instead of `end - start + 1` bytes after resuming from interruption even when start is specified. 

The same issue also probably happens [here](https://github.com/googleapis/google-resumable-media-python/blob/5d2091d78fb71e0d63254c5b731afb05af432104/google/resumable_media/requests/download.py#L223) in `Downloader`.

We encountered this issue when using the [google python storage](https://github.com/googleapis/python-storage) library in the setting where we had thousands of workers doing repeated range reads on the same GCS file. Code would run and return the wrong bytes somewhere 20-60min in. Adding in a fix (see pull request) seemed to resolve the issue.

wilson1yan:
Fixes #446

### DIFF
--- a/google/resumable_media/requests/download.py
+++ b/google/resumable_media/requests/download.py
@@ -220,8 +220,11 @@ class Download(_request_helpers.RequestsMixin, _download.Download):
             # To restart an interrupted download, read from the offset of last byte
             # received using a range request, and set object generation query param.
             if self._bytes_downloaded > 0:
+                start = self._bytes_downloaded
+                if self.start is not None:
+                    start += self.start
                 _download.add_bytes_range(
-                    self._bytes_downloaded, self.end, self._headers
+                    start, self.end, self._headers
                 )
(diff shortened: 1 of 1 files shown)


andrewsg:
Fixed in #450 to include regression tests. Thanks very much for your suggestion.

